PR #2, opened 2017-09-06
[DURACI-7300]Changed handling of events from que to multithreading

AndriyZabavskyy:
This is done to achieve higher quantity of processed message in time.
Implemented lock of DB, using findAndModify method in javadriver,
using method lockDocument in ObjectHandler. If for some reason event
did not lock the database, it will be added to waitlist. Has an issue in waitlist that events
are not published in Rabbitmq. Later today will try to implement comments from Daniel


AndriyZabavskyy:
Created another pull request without merge conflicts

PR #5, opened 2017-09-13
Changed handling of events from que to multithreading

AndriyZabavskyy:
This is done to achieve higher quantity of processed message in time.
Implemented lock of DB, using findAndModify method in javadriver,
using method lockDocument in ObjectHandler. If for some reason event
did not lock the database, it will be sended to waitlist


src/main/java/com/ericsson/ei/handlers/ObjectHandler.java
```diff
@@ -138,4 +139,28 @@ public JsonNode getAggregatedObject(String dbDocument) {
     public String extractObjectId(JsonNode aggregatedDbObject) {
         return aggregatedDbObject.get("_id").asText();
     }
+
```
evasiba-ericsson: add some comment here

src/main/java/com/ericsson/ei/mongodbhandler/MongoDBHandler.java
```diff
@@ -127,6 +129,20 @@ public  boolean updateDocument(String dataBaseName, String collectionName, Strin
         return false;
     }
```
evasiba-ericsson: Some description here too about what the method does and why we need it (because it finds and modifies record in one query).

src/main/java/com/ericsson/ei/mongodbhandler/MongoDBHandler.java
```diff
@@ -127,6 +130,20 @@ public  boolean updateDocument(String dataBaseName, String collectionName, Strin
...
 
+    public  DBObject findAndModify(String dataBaseName, String collectionName, String input, String updateInput){
```
evasiba-ericsson: methond findAndModify has still missing comment

PR #6, opened 2017-09-18
Implement Historical ER Query Service

Jibandeep:


src/main/java/com/ericsson/ei/erqueryservice/ERQueryService.java
```diff
@@ -0,0 +1,110 @@
...
+     * ER2.0 based on the eventID.
+     * 
+     * @param eventId
+     * @param searchParameters
+     * @param limitParam
+     * @param levelsParam
+     * @param tree
+     * @return ResponseEntity
+     */
+
```
evasiba-ericsson: We need to have the possibility to only fetch the upstream or only downstream or both.

Right now we are forced to fetch both upstream and downstream.

src/main/java/com/ericsson/ei/erqueryservice/ERQueryService.java
```diff
@@ -0,0 +1,110 @@
...
+    public ResponseEntity getEventSteamDataById(String eventId, JsonNode searchParameters, int limitParam,
```
evasiba-ericsson: Where is this method used?
Nothing testing it either.

src/test/java/com/ericsson/ei/erqueryservice/test/ERQueryServiceTest.java
```diff
@@ -0,0 +1,58 @@
...
+import java.net.URISyntaxException;
+import org.junit.Test;
+import org.springframework.http.HttpRequest;
+import org.springframework.http.server.ServletServerHttpRequest;
+import org.springframework.mock.web.MockHttpServletRequest;
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
+import org.springframework.web.util.UriComponents;
+import org.springframework.web.util.UriComponentsBuilder;
+
```
evasiba-ericsson: What are you testing here?

I do not see that you test the methods in ERQueryService.

PR #7, opened 2017-09-18
Implement SubscriptionHandler

Jibandeep:


src/main/java/com/ericsson/ei/subscriptionhandler/SubscriptionHandler.java
```diff
@@ -0,0 +1,158 @@
...
+    @Autowired
+    private MongoDBHandler handler;
+
+    @Autowired
+    private RunSubscription runSubscription;
+
+    @Autowired
+    private JmesPathInterface jmespath;
+
+    private JsonNode requirement;
```
evasiba-ericsson: What is this?

You cannot have private variables that can be set by some method and used by other method. You need to pass it as a result from methods.

src/main/java/com/ericsson/ei/subscriptionhandler/InformSubscription.java
```diff
@@ -0,0 +1,114 @@
...
+    SpringRestTemplate restTemplate;
+
+    @Autowired
+    MongoDBHandler mongoHandler;
+
+    @Autowired
+    SendMail sendMail;
+
+    static Logger log = (Logger) LoggerFactory.getLogger(InformSubscription.class);
+    static boolean saved = false;
```
evasiba-ericsson: Why this? The first document will be saved but all the others afterwards will not be saved since saved will always be true after first save.

This issue requires that you implement a new test case where an aggregation object misses to notify subscriber and then another one does it too so you should end up with two missed notifications in the database.

src/main/java/com/ericsson/ei/subscriptionhandler/SpringRestTemplate.java
```diff
@@ -0,0 +1,77 @@
...
+    public int postData(String aggregatedObject, String notificationMeta) {
+        JsonNode aggregatedJson = null;
+        ResponseEntity<JsonNode> response = null;
+        try {
+            aggregatedJson = new ObjectMapper().readTree(aggregatedObject);
+            response = rest.postForEntity(notificationMeta, aggregatedJson, JsonNode.class);
+        } catch (HttpClientErrorException e) {
+            log.error(e.getMessage(), e);
+            return HttpStatus.NOT_FOUND.value();
+        } catch (Exception e) {
```
evasiba-ericsson: Remove the above catch for HttpClientErrorException since is the same like this catch.

src/main/java/com/ericsson/ei/subscriptionhandler/InformSubscription.java
```diff
@@ -0,0 +1,114 @@
...
+            log.info("Notification through REST_POST");
+            int result = restTemplate.postData(aggregatedObject, notificationMeta);
+            if (result == HttpStatus.OK.value()) {
+                log.info("The result is : " + result);
+            } else {
+                for (int i = 0; i < failAttempt; i++) {
+                    try {
+                        log.info("Waiting for 2 seconds");
+                        Thread.currentThread().sleep(2000);
+                    } catch (InterruptedException e) {
```
evasiba-ericsson: This is a general catch so change it to catch(Exception

src/test/java/com/ericsson/ei/subscriptionhandler/test/SubscriptionHandlerTest.java
```diff
@@ -0,0 +1,103 @@
...
+    private SubscriptionHandler handler;
+
+    @Before
+    public void setUp() {
+
+        springRestTemplate = new SpringRestTemplate(new RestTemplateBuilder());
+        try {
+            aggregatedObject = FileUtils.readFileToString(new File(aggregatedPath));
+            subscriptionData = FileUtils.readFileToString(new File(subscriptionPath));
+        } catch (IOException e) {
```
evasiba-ericsson: change to catch(Exception

src/main/java/com/ericsson/ei/subscriptionhandler/RunSubscription.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * This method matches every condition specified in the subscription Object
+     * and if all conditions are matched then only the aggregatedObject is
+     * eligible for notification via e-mail or REST POST.
+     * 
+     * @param aggregatedObject
+     * @param requirement
+     * @param subscriptionJson
+     * @return boolean
+     */
```
evasiba-ericsson: There is a new test case you need to implement since I see a problem with your code. At firs subscription that fails it will stop and not check all the other subscriptions.

When we have an aggregated object and a set of three subscription. First subscription will fail the second will pass and the third will fail. Be sure that all the subscriptions are checked and not only the first one.

The second test is when you have a set of three subscriptions and only the first two pass. All three subscriptions needs to be run. 

Do not loop your self through the subscription set but you need to pass them to the subscriptionHandler so that as much of your code is tested.

src/main/java/com/ericsson/ei/subscriptionhandler/SpringRestTemplate.java
```diff
@@ -0,0 +1,77 @@
...
+            log.error(e.getMessage(), e);
+            return HttpStatus.NOT_FOUND.value();
+        }
+        HttpStatus status = response.getStatusCode();
+        log.info("The response code after POST is : " + status);
+        if (status == HttpStatus.OK) {
+            JsonNode restCall = response.getBody();
+            log.info("The response Body is : " + restCall);
+            return response.getStatusCode().value();
```
evasiba-ericsson: The return syntax is the same so you need to remove the else and the return from if statement. just one final return.

src/main/java/com/ericsson/ei/subscriptionhandler/InformSubscription.java
```diff
@@ -0,0 +1,114 @@
...
+ * This class represents the REST POST notification mechanism and the alternate
+ * way to save the aggregatedObject details in the database when the
+ * notification fails.
+ * 
+ * @author xjibbal
+ * 
+ */
+
+@Component
+public class InformSubscription {
```
evasiba-ericsson: Change name to InformSubscriber since it does not sound logical to inform a subscription.

PR #23, opened 2017-10-24
subscription handling updated & activated, clean up &  bugfixes

emichaf:


src/main/java/com/ericsson/ei/repository/SubscriptionRepository.java
```diff
@@ -11,10 +11,10 @@
 @Component
 public class SubscriptionRepository implements ISubscriptionRepository {
     
-    @Value("${subscription.collection.name}")
+    @Value("${subscriptionCollectionName}")
     private String collectionName;
     
-    @Value("${database.name}")
+    @Value("${subscriptionDataBaseName}")
```
evasiba-ericsson: database name is for all collections not only for subscription
Camelcase is not used in properties. Standard is point delimited.

src/main/resources/application.properties
```diff
@@ -28,10 +30,7 @@ waitlist.collection.name: wait_list
 waitlist.collection.ttlValue: 600
 waitlist.initialDelayResend: 2000
 waitlist.fixedRateResend: 15000
-database.name: eiffel_intelligence_eiffelxxx
-subscription.collection.name: subscription
-
-server.port: 8090
```
evasiba-ericsson: As I mentioned above these properties should be here as they are. Since the component will not find right properties to inject and camel case is not for use in properties naming.

emichaf: 2 different DBs and collection are used in the EI backend for subscription CRUD & Subscription handling. So updated are made in DB eiffel_intelligence_eiffelxxx collection: subscription and subscription handling is performed in DB: subscription and collection: subscription. Will remove DB subscription with collection:subscription. And update subscription handler to DB:  database.name & subscription.collection.name. Will remove subscriptionDataBaseName & subscriptionCollectionName

PR #26, opened 2017-10-26
copyright header added to test, and unit test updated

emichaf:
copyright header added to test, and unit test updated


emichaf:
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

The job has been terminated

emichaf:
Eventhandler updates for DownstreamIdRulesHandler kills incoming traffic, and breaks unit tests

PR #31, opened 2017-10-27
Added SubscriptionValidator class and unit tests. Add validation of n…

tobiasake:
…ew and updates of subscription before subscription is written to Database.

Now its ok to do the review and merge if the code changes is +1.

Good starting point is to review the unit test class and see if I validate test all situations. I have maybe missed some situation/usage of how each parameter is used.

This implementation works with EI Frontend and if a parameter fails in the validation, user is informed via a popup window in EI Frontend.

If you think we cover all situations in thes unit tests and in the validator class. Then give a review comments, example "Looks good." / "+1".

I will improve Javadoc etc. in next commit and if you feedback, I will take car of those at same time.
And after next commit we can merge this change.

### DIFF
--- a/src/main/java/com/ericsson/ei/subscriptionhandler/SubscriptionValidator.java
+++ b/src/main/java/com/ericsson/ei/subscriptionhandler/SubscriptionValidator.java
@@ -1,0 +1,97 @@
+package com.ericsson.ei.subscriptionhandler;
+
+//import java.util.List;
+import java.util.regex.Pattern;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.ericsson.ei.controller.model.Subscription;
+import com.ericsson.ei.exception.SubscriptionValidationException;
+//import com.ericsson.ei.controller.model.Requirement;
+
+public class SubscriptionValidator {
+	
+	private static final Logger LOG = LoggerFactory.getLogger(SubscriptionValidator.class);
+	
+	/*
+	 * Validation of parameters values in subscriptions objects.
+	 */
+	public void validateSubscription(Subscription subscription) throws SubscriptionValidationException {
+		
+		LOG.info("Validation of subscription " + subscription.getSubscriptionName() + " Started.");
+		
+		this.validateSubscriptionName(subscription.getSubscriptionName());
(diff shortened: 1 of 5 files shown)


src/main/java/com/ericsson/ei/subscriptionhandler/SubscriptionValidator.java
```diff
@@ -0,0 +1,137 @@
...
+		
+		this.validateSubscriptionName(subscription.getSubscriptionName());
+		this.validateNotificationMessage(subscription.getNotificationMessage());
+		this.validateNotificationMeta(subscription.getNotificationMeta());
+		this.validateNotificationType(subscription.getNotificationType());
+		this.validateRepeat(subscription.getRepeat());
+//		List<Requirement> reqList = subscription.getRequirements();
+//		for (int i=0; i < reqList.size(); i++) {
+//			this.validateJmespath(reqList.get(i).getConditions().get(0).getJmespath());
+//		}
```
tobiasake: I have comment out Jmepath part, since it do not validate anything yet. Need to find a good way to validate JmePath syntax. This can be done in a new story. Shall I remove the commented out Jmepath code or keep it comment out until JmePath syntax validator is implemented?

src/main/java/com/ericsson/ei/subscriptionhandler/SubscriptionValidator.java
```diff
@@ -0,0 +1,137 @@
...
+	 * format of parameter.
+	 * 
+	 * @param repeat
+	 */
+	public void validateRepeat(Boolean repeat) throws SubscriptionValidationException {
+		if (!(repeat instanceof Boolean)) {
+			throw new SubscriptionValidationException("Wrong format of Repeat: " + repeat.toString());
+		}
+		
+	}
```
tobiasake: Shall I remove this validator. It will never fail.
I guess it should fail earlier when EI Backend receives the EI RestApi request and decode it from JSON to Subscription Class, to the boolean repeat parameter/value. In this code it will always be boolean and never fail.
I feels like this check is unnecessary and can be removed.

tobiasake:
Removed the validation of Repeat flag in the Subscription object, since it will always be boolean and there is nothing to test.

PR #32, opened 2017-11-01
ADD: backend instance information

ned29:


src/main/java/com/ericsson/ei/controller/model/ParseInstanceInfoEI.java
```diff
@@ -0,0 +1,92 @@
...
+import com.ericsson.ei.waitlist.WaitListStorageHandler;
+import lombok.Getter;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+
+@Component
+public class ParseInstanceInfoEI {
```
AndriyZabavskyy: It would be helpful to add here comments that if some value parameters need to be in output of the JSON they should be added here.

src/main/resources/application.properties
```diff
@@ -1,6 +1,8 @@
 spring.application.name=eiffel-intelligence
 #${project.artifactId}
 
+build.version=@project.version@
+
```
tobiasake: Build version is not a setting option info.
Can "build.version" be put in a different file, that can handle these kind information?

tobiasake:
Looks ok,, but build.version should not be in application porperties file.
Another thing,, we discussed about this parameter "  spring.application.name=eiffel-intelligence " in properties file,, what did we decide,, was it to have in porperties file?

tobiasake:
I chatted with emichaf now, and its ok to have Version in application properties file.

tobiasake:
Only thing left to do, before merge, is to add RAML for the new InstanceInfoController. Will you do that? @ned29

src/main/resources/application.properties
tobiasake: Its ok to keep the version parameter here.

src/main/resources/public/raml/subscription.raml
```diff
@@ -80,4 +81,17 @@
             body:
               application/json:
                 schema: subscriptionResponse
-        
\ No newline at end of file
+
+title: Instance information
```
tobiasake: Indentation looks to not be correct here. Or the indentation is wrong in the beginning of the file.

PR #51, opened 2018-01-14
Fixed AMQP SSL/TLS connection between EI and RabbitMq.

tobiasake:


MagnusAronsson:
I've tried the patch out and it successfully establishes a secure connection to the message bus. It also still works fine with normal amqp. +1

PR #54, opened 2018-01-30
rules folder added with rulesets

emichaf:
rules folder added with rulesets


tobiasake:
Tests fails,,, Looks like execution has been interrupted or something? or?  Can't find any test failures in the log.

tobiasake:
Test execution ends in timeout due to a test seems to expect some events that never comes/is generated and test is waits forever for these events that never received. Tests(Test Framework) should fail after a shorter timeout if not all expected events is not received and produce a test failure report with info that these expected events was never received and these test failed due the issue.

PR #64, opened 2018-03-08
Subscription date handling moved to backend (epoch), multi subscripti…

emichaf:
Subscription date handling moved to backend (epoch), multi subscription upload support added. Flow tests updated/added.


src/main/java/com/ericsson/ei/controller/SubscriptionControllerImpl.java
```diff
@@ -105,12 +111,14 @@
 
     // @CrossOrigin
```
tobiasake: Should this be here? Remove if it is not needed/used.

PR #66, opened 2018-03-16
ADD: subscription template

ned29:


src/main/resources/subscriptionsTemplate.json
```diff
@@ -0,0 +1,58 @@
+[{
```
tobiasake: This file is already in the project file structure: src/test/resources/subscription.json.
So now we have duplicates, since this has been merged. Updates so we have only one subscription template file.

emichaf: I will remove it in my coming PR today!

PR #68, opened 2018-03-20
Rest end point for run rules on list events and find aggregated object

xmairaa:
- Rest end point for run rules on list events and find aggregated object
- Implemented code for multiple users access EI functionality.


src/main/java/com/ericsson/ei/services/RuleCheckService.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(RuleCheckService.class);
+
+    @Override
+    public String prepareAggregatedObject(String listRulesJson, String listEventsJson)
+            throws JSONException, JsonProcessingException, IOException {
+        JSONArray jsonObj = new JSONArray(listRulesJson);
+        JSONArray jsonObj2 = new JSONArray(listEventsJson);
+        eventHandler.getRulesHandler().setParsedJason(jsonObj.toString());
+        String aggregatedObjectId = null;
```
evasiba-ericsson: aggregatedObjectId needs to be a list since the action might result in several aggregated objects.

src/main/resources/application.properties
```diff
@@ -59,3 +59,6 @@ spring.mail.properties.mail.smtp.auth: false
 #spring.mail.properties.mail.smtp.starttls.enable: false
 
 er.url: http://localhost:8080/search/
+
+#Please disable this property in production environment, This /rules/rule-check/aggregation endpoint should work in test environment only
+#testaggregated.enabled: true
```
evasiba-ericsson: this property should be default false. Change the comment to Please keep this property value to false in production environment.

src/main/java/com/ericsson/ei/config/scope/register/CustomScopeRegisteringBeanFactoryPostProcessor.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.config.BeanFactoryPostProcessor;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.context.support.SimpleThreadScope;
+
+/**
+ * This class for the extend the request bean scope to thread level
+ * This scope is implemented to EventHandler and RuleHandler
+ * These two classes are not share to each request and rabbitmq consumer.
```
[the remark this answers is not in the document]
xmairaa: This doc from the previous commit, 
https://github.com/Ericsson/eiffel-intelligence/pull/68/files#diff-a7c593acd71e5cf31f200e0eb6789aff here i already modified doc

src/main/java/com/ericsson/ei/services/RuleCheckService.java
xmairaa: in our scenario (List of events & List of rules) gives only one aggregated object, ofcourse it is in List.

src/main/resources/application.properties
xmairaa: if we are not provide this property, it consider as false in production env, So No need in production.
I will modify comment lines like below

# Please enable this property in test environment only, This /rules/rule-check/aggregation endpoint should work in test environment only
# In production environment remove this property or false the value
#testaggregated.enabled: true

src/main/resources/application.properties
```diff
@@ -59,3 +59,7 @@ spring.mail.properties.mail.smtp.auth: false
 #spring.mail.properties.mail.smtp.starttls.enable: false
 
 er.url: http://localhost:8080/search/
+
+# Please enable this property in test environment only, This /rules/rule-check/aggregation endpoint should work in test environment only
+# In production environment remove this property or false the value
+testaggregated.enabled: true
```
evasiba-ericsson: this should be false even in the pull request so that you do not accidentally merge with value true

src/main/resources/application.properties
```diff
@@ -59,3 +59,7 @@ spring.mail.properties.mail.smtp.auth: false
 #spring.mail.properties.mail.smtp.starttls.enable: false
 
 er.url: http://localhost:8080/search/
+
+# Please enable this property in test environment only, This /rules/rule-check/aggregation endpoint should work in test environment only
+# In production environment remove this property or false the value
+#testaggregated.enabled: false
```
evasiba-ericsson: uncomment this property but let it set to false

src/main/java/com/ericsson/ei/handlers/EventToObjectMapHandler.java
```diff
@@ -135,6 +135,19 @@ public String getEventId(RulesObject rulesObject, String event) {
...
+    /**
+     * The method is responsible for the delete the EventObjectMap by using the suffix template Name
+     * 
+     * @param templateName
+     * @return boolean
+     */
+    public boolean deleteEventObjectMap(String templateName) {
+        String condition = "{\"objects\": { \"$in\" : [/.*" + templateName + "/]} }";
+        log.info("The Json condition for delete aggregated object is : " + condition);
+        boolean response = mongodbhandler.dropDocument(databaseName, collectionName, condition);
```
tobiasake: response can be removed and return directly dropDocument(...) return value.

src/main/java/com/ericsson/ei/queryservice/ProcessAggregatedObject.java
```diff
@@ -57,23 +57,50 @@
...
+     * the ID from the aggregatedObject.
+     * 
+     * @param id
+     * @return ArrayList
+     */
+    public ArrayList<String> getAggregatedObjectByTemplateName(String templateName) {
+        String condition = "{\"_id\": /.*" + templateName + "/}";
+        LOGGER.debug("The Json condition is : " + condition);
+        ArrayList<String> response = handler.find(aggregationDataBaseName, aggregationCollectionName, condition);
+        return response;
```
tobiasake: Same here. Or are we doing this extra step to show more clearly what type that is returned?

xmairaa:
Yes its good to have test report, But in this pull request we cont continue, We can implement those reports in separate pull request.

PR #70, opened 2018-03-23
UPDATE: source_change_object

ned29:


AndriyZabavskyy:
Looks good

src/test/java/com/ericsson/ei/flowtests/FlowSourceChangeObject.java
```diff
@@ -53,19 +53,24 @@ protected void setSpecificTestCaseParameters() {
...
     protected void checkResult() {
         try {
-            String expectedDocuments = FileUtils.readFileToString(new File(inputFilePath));
+            String expectedDocuments = FileUtils.readFileAsString(new File(inputFilePath));
             ObjectMapper objectmapper = new ObjectMapper();
             JsonNode expectedJsons = objectmapper.readTree(expectedDocuments);
             String document = objectHandler.findObjectById("fb6efi4n-25fb-4d77-b9fd-5f2xrrefe66de47");
             JsonNode actualJson = objectmapper.readTree(document);
-            assertEquals(expectedJsons.toString().length(), actualJson.toString().length(),9);
+            assertEquals(expectedJsons.toString().length(), actualJson.toString().length());
```
tobiasake: Is it possible to get a report of which event have failed and succeded in the test and not just compare Json content.
Its ok to compare json content, but it should be clear what failed ,,, e.g. which event was missed.

ned29: if some event missing than test will failed, because they length will not equals

tobiasake: Yes, but the length doesn't say what failed or what is missing. You just compare the objects has same length -> e.g. same content i best case.
This  is a general comment for all test in EI,, especially FlowTests in EI, where it is very hard to say what failed when it fails. My opinion is that all tests that handle events and aggregate events data should produce a test result report that clearly informs what events and aggregated objects has expected content and if an aggregated object or events is missing, it should clearly show what event/data is missing.
Length of a JSON object doesn't say anything or help in debugging work.
It should be clear what failed and what is missing (Events, parameters in JSON, etc.).

ned29: fixed, if some event is missing will write which field not found

PR #77, opened 2018-04-01
Refactored code in flow tests, fixed generating of TestExecution object

ValentinTyhonov:
Refactored code of flow tests
Removed few unused files
Fixed using deprecated method

Fixed few steps in generating TestExecution object:
- moved test_suite_name to test_suite array
- added dependency between events - TestSuiteStarted <- TestCaseTriggered <- TestCaseStarted <- TestCaseFinished instead of ActivityTriggered <- TestSuiteStarted, TestCaseTriggered, TestCaseStarted, TestCaseFinished (fixed issue that sometimes "ongoing" value was incorrect)


src/test/java/com/ericsson/ei/flowtests/FlowTestExternalComposition.java
```diff
@@ -16,55 +16,50 @@
...
-                JsonNode actualJson3 = objectmapper.readTree(document3);
-                int i = 0;
-            } catch (Exception e) {
-                log.info(e.getMessage(),e);
-            }
+    @Override
+    List<String> setEventNamesToSend() {
+        List<String> eventNames = new ArrayList<>();
+        eventNames.add("event_EiffelArtifactCreatedEvent_3");
+        //eventNames.add("event_EiffelCompositionDefinedEvent_4");
```
tobiasake: Will these event not be used?

ValentinTyhonov: This test was not workable.. and part of events were commented. so I left them for future

tobiasake: Ok,, checked the git log and those were commented out since earlier. So leave it as is.

PR #86, opened 2018-04-17
Add LDAP authorization and MongoDB session storage

Christoffer-Cortes:


src/main/java/com/ericsson/ei/config/HttpSessionConfig.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.session.data.mongo.JdkMongoSessionConverter;
+import org.springframework.session.data.mongo.config.annotation.web.http.EnableMongoHttpSession;
+
+@EnableMongoHttpSession(maxInactiveIntervalInSeconds = 1200, collectionName="sessions")
+public class HttpSessionConfig {
+
+    @Bean
+    public JdkMongoSessionConverter jdkMongoSessionConverter() {
+        return new JdkMongoSessionConverter(Duration.ofMinutes(30));
```
ValentinTyhonov: Shouldn't these values (1200  and 30) be set from properties file?

Christoffer-Cortes: The one in JdkMongoSessionConverter could be but not the ones in EnableMongoHttpSession annotation. It does not support getting values from properties. If i type "${my.expiration.interval}" it will not accept a String and if i convert it to int using Integer.parseInt() it will tell me that it must be a constant expression. So it appears the annotation does not support getting values from properties. I don't know how to work around this. The value in JdkMongoSessionConverter is called maxInactiveInterval just like the one in annotation but I'm not sure what it does since the parameter that controls the session expiration is the one in annotation.

evasiba-ericsson: It can be done in similar way as here:
https://github.com/spring-projects/spring-session/issues/114

pom.xml
```diff
@@ -10,7 +10,7 @@
     <parent>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-starter-parent</artifactId>
-        <version>1.5.5.RELEASE</version>
+        <version>2.0.1.RELEASE</version>
```
tobiasake: Have we discussed and decided on lifting all Eiffel components to the new Spring 2.0 version?

src/main/java/com/ericsson/ei/EndpointSecurity.java
```diff
@@ -0,0 +1,89 @@
...
+            auth
+            .eraseCredentials(false)
+            .ldapAuthentication()
+                .userSearchFilter(ldapUserFilter)
+                .contextSource()
+                    .url(ldapUrl)
+                    .root(ldapBaseDn)
+                    .managerDn(ldapUsername)
+                    .managerPassword(decodeBase64(ldapPassword));
+        }
```
tobiasake: Maybe we should add log a message here which says it configured with LDAP or Not using LDAP. Good to have when debugging in logs.

src/main/java/com/ericsson/ei/EndpointSecurity.java
```diff
@@ -0,0 +1,89 @@
...
+            .and()
+                .httpBasic()
+            .and()
+                .csrf().disable();
+        }
+        else {
+            http
+            .csrf().disable();
+        }
+    }
```
tobiasake: Maybe we should add log a message here which says it configured with LDAP or Not using LDAP. Good to have when debugging in logs.

src/main/java/com/ericsson/ei/EndpointSecurity.java
```diff
@@ -59,6 +59,7 @@ protected void configure(HttpSecurity http) throws Exception {
             LOGGER.info("LDAP security configuration is enabled");
             http
             .authorizeRequests()
+                .antMatchers("/login/auth/").permitAll()
```
ValentinTyhonov: endpoint is /auth/login, isn't it?

Christoffer-Cortes: Yes, my misstake.

PR #87, opened 2018-04-23
Added controller for authorization

ValentinTyhonov:
This endpoint is used for getting authorization request from frontend


src/test/java/com/ericsson/ei/controller/TestLoginControllerImpl.java
```diff
@@ -0,0 +1,33 @@
...
+    private MockMvc mockMvc;
+
+    @MockBean
+    private LoginController loginController;
+
+    @Test
+    public void testResponseStatus() throws Exception {
+        mockMvc.perform(MockMvcRequestBuilders.get("/auth/login")
+            .accept(MediaType.APPLICATION_JSON_VALUE))
+            .andExpect(status().isOk())
```
xmairaa: we should check response body also.

src/main/java/com/ericsson/ei/controller/LoginControllerImpl.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    @CrossOrigin
+    @ApiOperation(value = "To login user", response = String.class)
+    public ResponseEntity<?> getAuthLogin() {
+        try {
+            String currentUser = SecurityContextHolder.getContext().getAuthentication().getName();
+            JSONObject json = new JSONObject();
+            json.put("user", currentUser);
+            return new ResponseEntity<>(json.toString(), HttpStatus.OK);
```
xmairaa: instead of initializing json object we can manage with string Like **"{'user':'"+currentUser+"'}"**

src/main/resources/public/raml/auth.raml
```diff
@@ -0,0 +1,8 @@
+/login:
+  get:
+    description: This call for login user to EI
```
xmairaa: This call for getting the logged in user, So change content according.

PR #92, opened 2018-04-23
Handling authorization in Subscription REST API

sericcson:


src/main/java/com/ericsson/ei/controller/SubscriptionControllerImpl.java
```diff
@@ -180,4 +198,10 @@
             return new ResponseEntity<List<Subscription>>(subscriptionList, HttpStatus.OK);
         }
     }
+
+    public String currentUser() {
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+        user = authentication.getName();
```
tobiasake: you can return autgentication.getName() directly and skip the user variable.

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -56,9 +58,17 @@ public boolean addSubscription(Subscription subscription) {
...
+    public Subscription getSubscription(String name, String userName) throws SubscriptionNotFoundException {
         
         String query = String.format(SUBSCRIPTION_NAME, name);
+
+        if(!userName.isEmpty()){  
+            String queryUser = String.format(USER_NAME, userName);      
+            String operatorAnd = "{$and:[%s]}";
+            String queryTemp = query+ "," + queryUser;            
+            query = String.format(operatorAnd, queryTemp);            
+        }
```
tobiasake: I don't understand why you need to query for both SUBSCRIPTION_NAME and USER_NAME in same query. It can only exist one 'SUBSCRIPTION_NAME' and adding USER_NAME to the query will make same result as query only SUBSCRIPTION_NAME.
Or have I miss-understood the functionality in this function?
After it have found the SUBSCRIPTION_NAME document,, then you can validate if the subscription document belongs to the current user.

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -83,8 +90,15 @@ public Subscription getSubscription(String name) throws SubscriptionNotFoundExce
...
     @Override
-    public boolean doSubscriptionExist(String name) {
+    public boolean doSubscriptionExist(String name, String userName) {
         String query = String.format(SUBSCRIPTION_NAME, name);
+
+        if(!userName.isEmpty()){  
+            String queryUser = String.format(USER_NAME, userName);      
+            String operatorAnd = "{$and:[%s]}";
+            String queryTemp = query+ "," + queryUser;            
+            query = String.format(operatorAnd, queryTemp);
```
tobiasake: Same comment here.

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -93,26 +107,43 @@ public boolean doSubscriptionExist(String name) {
...
         ObjectMapper mapper = new ObjectMapper();
         try {
             String StringSubscription = mapper.writeValueAsString(subscription);
             String subscriptionNameQuery = String.format(SUBSCRIPTION_NAME, subscriptionName);
+ 
+            if(!userName.isEmpty()){  
+                String queryUser = String.format(USER_NAME, userName);      
+                String operatorAnd = "{$and:[%s]}";
+                String queryTemp = subscriptionNameQuery+ "," + queryUser;            
+                subscriptionNameQuery = String.format(operatorAnd, queryTemp);
```
tobiasake: Same comment here also.

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -93,26 +107,43 @@ public boolean doSubscriptionExist(String name) {
...
     @Override
-    public boolean deleteSubscription(String name) {
+    public boolean deleteSubscription(String name, String userName) {
         String query = String.format(SUBSCRIPTION_NAME, name);
+        if(!userName.isEmpty()){  
+            String queryUser = String.format(USER_NAME, userName);      
+            String operatorAnd = "{$and:[%s]}";
+            String queryTemp = query+ "," + queryUser;            
+            query = String.format(operatorAnd, queryTemp);            
+        }
```
tobiasake: same comment here as well.
And it looks like to be exactly same code in all place. Break it out to own function and reuse the code where its needed in the code.

src/main/resources/subscriptionsTemplate.json
```diff
@@ -19,7 +20,10 @@
                 }
             ]
         }
-    ]
+
+    ],
+    "subscriptionName" : "Subscription_Test",
```
tobiasake: Remove subscriptionName line here. You have already one subscriptionName line in the beginning:  "subscriptionName" : "Subscription1", ....

src/main/resources/subscriptionsTemplate.json
```diff
@@ -42,5 +46,8 @@
                 }
             ]
         }
-    ]
+
+    ],
+    "subscriptionName" : "Subscription_Test_Modify",
```
tobiasake: Remove subscriptionName line here. You have already one subscriptionName line in the beginning:  "subscriptionName" : "Subscription2" ....

src/main/java/com/ericsson/ei/controller/SubscriptionControllerImpl.java
```diff
@@ -45,20 +48,32 @@
...
+
     @Autowired
     private ISubscriptionService subscriptionService;
 
     private SubscriptionValidator subscriptionValidator = new SubscriptionValidator();
 
     private static final Logger LOG = LoggerFactory.getLogger(SubscriptionControllerImpl.class);
 
+    // Adding variable(for development purpose) until auth is implemented
+    private String user = "";
```
vasba: this should not be global. Add it as local variable in the methods where you use them. Remove the comment too.

src/main/resources/application.properties
```diff
@@ -61,6 +61,8 @@ email.subject: Email Subscription Notification
 notification.failAttempt: 3
 notification.ttl.value: 600
 
+ldap.enabled: true
```
vasba: this should be false by default . use it true only when you test locally

src/main/java/com/ericsson/ei/controller/SubscriptionControllerImpl.java
```diff
@@ -180,4 +198,11 @@
             return new ResponseEntity<List<Subscription>>(subscriptionList, HttpStatus.OK);
         }
     }
+
+    public String currentUser() {
+        // Authentication authentication =
+        // SecurityContextHolder.getContext().getAuthentication();
+        user = SecurityContextHolder.getContext().getAuthentication().getName();
+        return user;
```
ValentinTyhonov: no need to create variable. And remove comments

PR #94, opened 2018-04-28
Added new endpoints for checking status and security

ValentinTyhonov:
+ added endpoint for checking backend status
+ added endpoint for checking is backend secured or not
+ secured only needed endpoints
+ fixed logout
+ fixed modifying subscription, when security is enabled


src/main/java/com/ericsson/ei/controller/AuthControllerImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Value("${ldap.enabled:false}")
+    private boolean ldapEnabled;
+
+    @Override
+    @CrossOrigin
+    @ApiOperation(value = "To check is security enabled", response = String.class)
+    public ResponseEntity<?> getAuth() {
+        try {
+            return new ResponseEntity<>("{\"security\":\"" + ldapEnabled + "\"}", HttpStatus.OK);
```
tobiasake: Is these extra quotation needed "\"" in the response? Is it required by Js code to read Response in EI Frontend?

ValentinTyhonov: It is needed for correct parsing response in frontend

PR #98, opened 2018-05-16
Removed checking by userName from getSubscription method

ValentinTyhonov:
Removed checking by userName from getSubscription method -> now every user can view every subscription, but they still can remove or modify only own subscriptions.

Removed checking by userName from doExistSubscription method -> it was bad idea to check it be userName, because then user could create subscription with the SubscriptionName (id), even if other one is already creted with the same name  by other user.


src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -79,20 +81,18 @@ public Subscription getSubscription(String name, String userName) throws Subscri
                 return subscription;
                 // return mapper.readValue(input, Subscription.class);
```
tobiasake: Remove this out-commented line if its not going to be used.

src/main/java/com/ericsson/ei/services/ISubscriptionService.java
```diff
@@ -44,19 +44,21 @@
...
+    Subscription getSubscription(String name) throws SubscriptionNotFoundException;
     
     /**
      * 
      * @param subscription
      * @param subscriptionName
+     * @param user
      * @return
      */
     boolean modifySubscription(Subscription subscription, String subscriptionName, String user);
```
tobiasake: I don't understand why we need user as input. An existing Subscription should have userName field already. And subscriptionName is the key for the unique subscription.  Why do we need user parameter? It should be removed in my point of view. Or am I missing something?

tobiasake: Do we use 'user', username, in more places where its not needed?

ValentinTyhonov: here we use userName for preventing modify subscription if current user is not owner of this subscription.
only in two places user param was unneeded, so there I removed this param

tobiasake: Ok,, and the user value comes from the RestApi authentication mechanism. Then its ok.

PR #103, opened 2018-05-21
Ttl update for aggregation

sericcson:


src/main/resources/application.properties
```diff
@@ -60,6 +60,8 @@ email.subject: Email Subscription Notification
 notification.failAttempt: 3
 notification.ttl.value: 600
 
+ldap.enabled: false
```
ValentinTyhonov: Why you adding new ldap.enabled property if there is already one at the line 74?

src/main/java/com/ericsson/ei/handlers/ObjectHandler.java
```diff
@@ -218,4 +220,16 @@ public String lockDocument(String id) {
         }
         return null;
     }
+
+    public int getTtl() {
+        int ttl = 0;
+        if (ttlValue != null && !ttlValue.isEmpty()) {
```
ValentinTyhonov: You can use StringUtils.isNotBlank(ttlValue) instead

PR #108, opened 2018-05-31
Functional Tests Subscription CRUD

sericcson:


tobiasake:
Can someone explain to me what own implementation logic we are testing here?
To me these test looks like we are testing the external MockMVC library.
Why do we have these tests when its testing NO code?

PR #109, opened 2018-06-04
Added authentication to MongoDB

ValentinTyhonov:


src/main/java/com/ericsson/ei/mongodbhandler/MongoDBHandler.java
```diff
@@ -65,16 +60,35 @@
...
+    @Value("${spring.data.mongodb.database}")
+    private String database;
+
+    @Getter
+    @Value("${spring.data.mongodb.username}")
+    private String username;
+
+    @Getter
+    @Value("${spring.data.mongodb.password}")
+    private char[] password;
```
tobiasake: In Travis test it fails on that password can be set directly to char.
You need probably change it to string and check that if not null and not empty, then append password String to char array.

PR #116, opened 2018-06-12
Set Aggregated Rules RestApi default enabled

tobiasake:


tobiasake:
Ok, But the EI-Frontend need to be implement in the same way, -> Test Rules page disabled per default.
Its really strange to have it disabled in EI-Backend default and in same time have it enabled in EI-Frontend(Test Rules). And have it as today with strange Test Environment Error message which causes even more question marks. Either Disable it totally in EI-Frontend or have a really informative message on Test Rule page that his functionality is Disable and explain how to enable it. I had mail thread earlier earlier about this, but I got no answer on why it was disabled,, so I made this PR so its enabled as it is in EI-Frontend. We need to decide how we want this. As it is today is very wrong in my user perspective.

tobiasake:
I have changed so Test Rules RestApi setting to false and solved a issue in EI Instance info object and added some EI instance infos.

src/main/java/com/ericsson/ei/controller/model/ParseInstanceInfoEI.java
```diff
@@ -37,21 +36,26 @@
...
     @Getter
     @Value("${spring.application.name}")
     private String applicationName;
 
     @Getter
     @Value("${build.version}")
     private String version;
+    
+    @Getter
+    @Value("${testaggregated.enabled}")
```
ValentinTyhonov: you can set  default value for this property here (like ("${testaggregated.enabled:false}").  then even if this property is missed in application.properties, you will not get error.

src/main/java/com/ericsson/ei/controller/model/ParseInstanceInfoEI.java
```diff
@@ -80,6 +92,78 @@
...
+        @Getter
+        @Value("${ldap.user.filter}")
+        private String filter;
+        
+        @Getter
+        @Value("${ldap.username}")
+        private String username;
+
+        @Getter
+        private String password = "*********";
```
ValentinTyhonov: Is this field really needed in information in frontend?

PR #117, opened 2018-06-13
Added functional test for authentication

ValentinTyhonov:
The create subscription operation faisl when performed without credentials but it succeeds when performed with credentials.


src/functionaltests/resources/features/authentication.feature
```diff
@@ -0,0 +1,29 @@
...
+## (Comments)
+#Sample Feature Definition Template
+Feature: Test Authentication
+
+  Scenario: Call an REST API without credentials
+    Given LDAP is activated
+    When make a POST request to the subscription REST API "/subscriptions" without credentials
+    Then get response code of 401 and subscription with name "Subscription_Test" is not created
+
+	Scenario: Call an REST API with credentials
```
tobiasake: fix indentation.

src/functionaltests/resources/features/authentication.feature
```diff
@@ -0,0 +1,29 @@
...
+		Given LDAP is activated
+		When make a POST request to the subscription REST API "/subscriptions" with username "gauss" and password "password"
```
tobiasake: Maybe you should have SubscriptionName "Subscription_Test" mentioned as input in when step also. Or do you use the template file with the existing defined SubscriptionName.

src/functionaltests/java/com/ericsson/ei/utils/TestConfigs.java
```diff
@@ -76,6 +78,17 @@ MongoClient mongoClient() throws IOException {
...
 
+    @Bean
+    void setAuthorization() {
+        String password = StringUtils.newStringUtf8(Base64.encodeBase64("password".getBytes()));
+        System.setProperty("ldap.enabled", "true");
+        System.setProperty("ldap.url", "ldap://ldap.forumsys.com:389/dc=example,dc=com");
+        System.setProperty("ldap.base.dn", "dc=example,dc=com");
+        System.setProperty("ldap.username", "cn=read-only-admin,dc=example,dc=com");
+        System.setProperty("ldap.password", password);
+        System.setProperty("ldap.user.filter", "uid={0}");
```
tobiasake: Do you really need to set all these parameters? Can't you mock the Ldap lookup/call with just Access granted return value or what valid value that ldap returns when a user get autherized ?

src/functionaltests/resources/features/authentication.feature
a-breid-ericsson: Should it not be @tag1 @tag2 before the scenarios also? Thats how Vali did it.

ValentinTyhonov: I use template with existed subscriptionName. Actually I can try to implement name changing and pass subscriptionName as an input parameter if it is needed.

What about tags, not sure for what they are needed.

src/functionaltests/resources/features/authentication.feature
ValentinTyhonov: Ok.

src/functionaltests/java/com/ericsson/ei/utils/TestConfigs.java
ValentinTyhonov: For testing ldap, ldap.enabled should be set to true, If we set ldap.enabled: true, then all these parameters are required.

src/functionaltests/java/com/ericsson/ei/subscriptions/authentication/AuthenticationSteps.java
```diff
@@ -0,0 +1,103 @@
...
+    private String requestBody;
+
+    @Given("^LDAP is activated$")
+    public void ldap_is_activated() throws Throwable {
+        String responseBody = new JSONObject().put("security", true).toString();
+        mockMvc.perform(MockMvcRequestBuilders.get("/auth")
+            .accept(MediaType.APPLICATION_JSON_VALUE))
+            .andExpect(status().isOk())
+            .andExpect(content().string(responseBody))
+            .andReturn();
```
tobiasake: Do we need to log what failed or do mockMvc log a good detailed description on what failed when result is not as Expected ?

tobiasake: This comment applies to all test steps.

src/functionaltests/java/com/ericsson/ei/subscriptions/authentication/AuthenticationSteps.java
```diff
@@ -0,0 +1,103 @@
...
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.MvcResult;
+import org.springframework.test.web.servlet.request.MockMvcRequestBuilders;
+
+import java.io.File;
+
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@Ignore
```
tobiasake: Why ignore annotation here?

src/functionaltests/resources/features/authentication.feature
sericcson: @tag are commented... comment for the scenario

src/functionaltests/java/com/ericsson/ei/subscriptions/authentication/AuthenticationSteps.java
ValentinTyhonov: As you can see, previous build failed, after I removed ignore annotation. *Steps.class should be ignored, because there are no runnable methods, that causes errors

PR #124, opened 2018-06-26
 Fixed wait list creates performance issues

ValentinTyhonov:
* Added checking that the event we put in the wait list is already there and adding it only if it is not in wait list collection.
* Drop event from wait list if it is in event object map collection.


src/main/java/com/ericsson/ei/handlers/EventToObjectMapHandler.java
```diff
@@ -148,5 +149,11 @@ public boolean deleteEventObjectMap(String templateName) {
         return mongodbhandler.dropDocument(databaseName, collectionName, condition);
     }
 
+    public boolean isEventExistsInEventObjectMap(String eventId) {
```
evasiba-ericsson: Remove Exists from the name

src/main/java/com/ericsson/ei/waitlist/WaitListStorageHandler.java
```diff
@@ -62,46 +61,47 @@
...
+    private String addPropertiesToEvent(String event, RulesObject rulesObject) {
+        String time;
+        Date date = new Date();
         String idRule = rulesObject.getIdRule();
         JsonNode id = jmesPathInterface.runRuleOnEvent(idRule, event);
-        String condition = "{Event:" + JSON.parse(event).toString() + "}";
-        ArrayList<String> documents = mongoDbHandler.find(databaseName, collectionName, condition);
-        if (documents.size() == 0) {
+        String condition = "{Event:" + event + "}";
+        List<String> documents = mongoDbHandler.find(databaseName, collectionName, condition);
```
evasiba-ericsson: this query is no longer necessary.

PR #125, opened 2018-06-26
Improved responses

ValentinTyhonov:
Improved response messages for all REST API, where it was needed


src/main/java/com/ericsson/ei/controller/info/InstanceInfoControllerImpl.java
```diff
@@ -39,8 +39,9 @@ public String parseInfo() {
         try {
             return new ObjectMapper().writerWithDefaultPrettyPrinter().writeValueAsString(istanceInfo);
         } catch (Exception e) {
-            LOGGER.error("Serialization has failed " + e.getMessage());
+            String errorMessage = "Failed to parse EI backend information. Error message:\n" + e.getMessage();
+            LOGGER.error(errorMessage);
+            return errorMessage;
         }
```
tobiasake: This will cause that nothing will happen in EI Frontend,,, I guess just a white page._
I think this should be rewritten to use ResponseEntity. And update EI Frontend, if needed.

src/main/java/com/ericsson/ei/controller/AuthControllerImpl.java
```diff
@@ -74,7 +74,7 @@
     @ApiOperation(value = "To check backend status", response = String.class)
     public ResponseEntity<?> getCheckStatus() {
         try {
-            return new ResponseEntity<>("Backend is active", HttpStatus.OK);
+            return new ResponseEntity<>(HttpStatus.OK);
```
tobiasake: I think you can can keep a message to the Http OK code, e.g: "Backend server is up and running"

src/main/java/com/ericsson/ei/controller/info/InstanceInfoControllerImpl.java
ValentinTyhonov: in frontend we have white page and error message in jGrowl. I can print message in page instead of jGrowl

src/main/java/com/ericsson/ei/controller/AuthControllerImpl.java
ValentinTyhonov: I tested it with frontend and somehow EI Status Button is red if REST API returns some message

src/main/java/com/ericsson/ei/controller/info/InstanceInfoControllerImpl.java
tobiasake: ok,, jGrowl and sticky jGrowl is OK to use.
But how do Frontend JS code know if its http Error or not? when we don't have any HTTP Status code,, or am I missing something.

src/main/java/com/ericsson/ei/controller/AuthControllerImpl.java
tobiasake: Ok, aren't we just checking the status code in frontend? ,,, then the message should be "don't care" value.

ValentinTyhonov: Fixed. I have changed contentType for this ajax call in frontend

src/main/java/com/ericsson/ei/controller/info/InstanceInfoControllerImpl.java
ValentinTyhonov: Somehow JS know it. But if needed, I will use Response Entity.

tobiasake: I think it is more correct to use ResponseEntity as response. Then we also get same way of handling response in every entrypoint.

ValentinTyhonov: Done

src/main/java/com/ericsson/ei/controller/AuthControllerImpl.java
tobiasake: Great

PR #130, opened 2018-07-04
Add functional tests for time to live and notification retries

e-pettersson-ericsson:
Testing time to live on aggregated objects and missed notifications in database. 
Testing notification retries.


src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+        private static final String BASE_URL = "localhost";
+        private static final String ENDPOINT = "/missed_notification";
+
+        private static final String SUBSCRIPTION_FILE_PATH = "src/functionaltests/resources/SubscriptionObject.json";
+        private static final String AGGREGATED_OBJECT_FILE_PATH = "src/functionaltests/resources/AggregatedObject.json";
+        private static final String MISSED_NOTIFICATION_FILE_PATH = "src/functionaltests/resources/MissedNotification.json";
+
+        private static String subscriptionObject;
+        private static String aggregatedObject;
+        private static String missedNotification;
```
a-breid-ericsson: missedNotification is only used in one method and should maybe be a local variable in that method.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+        private static final Logger LOGGER = LoggerFactory.getLogger(TestTTLSteps.class);
+        private static final String BASE_URL = "localhost";
+        private static final String ENDPOINT = "/missed_notification";
+
+        private static final String SUBSCRIPTION_FILE_PATH = "src/functionaltests/resources/SubscriptionObject.json";
+        private static final String AGGREGATED_OBJECT_FILE_PATH = "src/functionaltests/resources/AggregatedObject.json";
+        private static final String MISSED_NOTIFICATION_FILE_PATH = "src/functionaltests/resources/MissedNotification.json";
+
+        private static String subscriptionObject;
+        private static String aggregatedObject;
```
a-breid-ericsson: aggregatedObject, although it is used in 2 methods, they are set individually in those methods with new values, maybe they should be a local variable in those methods.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+        public void afterScenario() throws IOException {
+            LOGGER.debug("Shutting down mock servers.");
+            mockServerClient.close();
+            clientAndServer.stop();
+        }
+
+        // START TEST SCENARIOS
+
+        @Given("^Missed notification is created in database with index \"([A-Za-z0-9_]+)\"$")
+        public void missed_notification_is_created_in_database(
```
a-breid-ericsson:
```
public void missed_notification_is_created_in_database(
    String indexName) throws IOException {
```

That new line seems not to be correct.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+                    indexName);
+
+            // setting 1 second TTL on index in db
+            mongoDBHandler.createTTLIndex(missedNotificationDatabase, missedNotificationCollectionName,
+                    indexName, 1);
+            Boolean isInserted = mongoDBHandler
+                    .insertDocument(missedNotificationDatabase,
+                            missedNotificationCollectionName,
+                            missedNotificationDocument.toString());
+            assertEquals(true, isInserted);
```
a-breid-ericsson: `assertEquals(true, isInserted);`

A comment if this fails would be great, as it is you will only see that something expected true but got false.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+
+            List<String> result = mongoDBHandler.getAllDocuments(missedNotificationDatabase,
+                    missedNotificationCollectionName);
+            assertEquals("Database " + missedNotificationDatabase + " is not empty.", false, result.isEmpty());
+        }
+
+        @Given("^Aggregated object is created in database with index \"([A-Za-z0-9_]+)\"$")
+        public void aggregated_object_is_created_in_database(
```
a-breid-ericsson:
```
public void aggregated_object_is_created_in_database(
    String indexName) throws IOException {
```
This line break seems also odd to me.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+            Boolean isInserted = mongoDBHandler.insertDocument(aggregationDataBaseName,
+                    aggregationCollectionName, aggregatedDocument.toString());
+            assertEquals(true, isInserted);
+
+            List<String> result = mongoDBHandler.getAllDocuments(aggregationDataBaseName,
+                    aggregationCollectionName);
+            assertEquals("Database " + aggregationDataBaseName + " is not empty.", false, result.isEmpty());
+        }
+
+        @When("^I sleep for \"([^\"]*)\" seconds")
```
a-breid-ericsson: What if instead of sleeping for 60 seconds always, rewrite the check in "xx has been deleted from yy" to check for 60 seconds? That way if it was deleted after 30 seconds the test could continue more quickly.

So the When could be "{DB-entry} has been deleted from {database} within {seconds}"

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+            // The background task deleting notifications from Mongo DB
+            // runs every 60 seconds
+            TimeUnit.SECONDS.sleep(sleepTime);
+        }
+
+        @Then("^\"([^\"]*)\" has been deleted from \"([A-Za-z0-9_]+)\" database$")
+        public void has_been_deleted_from_database(String collection, String database) {
+            LOGGER.debug("Checking " + collection + " in " + database);
+            List<String> result = mongoDBHandler.getAllDocuments(database, collection);
+            assertEquals("[]", result.toString());
```
a-breid-ericsson: Please add a comment to this asserEquals

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        // SCENARIO @TestNotificationRetries
+
+        @Given("^No missed notifications exists in database$")
```
a-breid-ericsson: @Given("^No missed notifications exists in database$
This has no assert, is it a test? maybe just to have this one line of code in the Given where collection is created?

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+        public void no_missed_notifications_exists_in_database() {
+            mongoDBHandler.dropCollection(missedNotificationDatabase, missedNotificationCollectionName);
+        }
+
+        @Given("^Subscription is created$")
+        public void create_subscription() throws IOException, JSONException {
+            LOGGER.debug("Starting scenario @TestNotificationRetries.");
+            subscriptionObject = FileUtils.readFileToString(new File(SUBSCRIPTION_FILE_PATH), "utf-8");
```
a-breid-ericsson: You should use "getJSONFromFile(filePath)" from the FunctionalTestBase instead of FileUtils.readFileToString(new File(SUBSCRIPTION_FILE_PATH), "utf-8");

This applies to all places in this code where you read a json-file.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+
+        // SCENARIO @TestNotificationRetries
+
+        @Given("^No missed notifications exists in database$")
+        public void no_missed_notifications_exists_in_database() {
+            mongoDBHandler.dropCollection(missedNotificationDatabase, missedNotificationCollectionName);
+        }
+
+        @Given("^Subscription is created$")
+        public void create_subscription() throws IOException, JSONException {
```
a-breid-ericsson: The name of this method indicates that you create a subscription, but you create a subscription object.
It seems to be preperation for a test, maybe those lines should be done right be fore the subscription object is used.

a-breid-ericsson: Okey, seems to be small test, but you may want to assert that port was successfully changed in the object..

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+
+        @Then("^Verify that request has been retried")
+        public void verify_request_has_been_made() throws JSONException {
+            String retrievedRequests = mockServerClient.retrieveRecordedRequests(request()
+                    .withPath(ENDPOINT), Format.JSON);
+            JSONArray requests = new JSONArray(retrievedRequests);
+
+            //TODO: wait for requests to have been made in travis
+            // received requests include number of retries
+            assertEquals(2, requests.length());
```
a-breid-ericsson: Plaease add a comment in this assertion :)

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,232 @@
...
+         */
+        private BasicDBObject prepareDocumentWithIndex(String fileContent, String fieldName) {
+            Date date = new Date();
+            DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
+            String time = dateFormat.format(date);
+            try {
+                    date = dateFormat.parse(time);
+            } catch (Exception e) {
+                    LOGGER.error(e.getMessage(), e);
+            }
```
a-breid-ericsson: You could make this one trow an exception instead of using try catch.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
tobiasake: Sleeping is that a step step?  I think this should be moved to the end of the previoues test step.
A test step should test something and do some assertion on some actual values and compare it with some expected values.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,225 @@
...
+
+        @When("^I sleep for \"([^\"]*)\" seconds")
+        public void i_sleep_for_time(int sleepTime) throws Throwable {
+            LOGGER.debug("Sleeping for " + sleepTime + " at " + LocalTime.now());
+            // The background task deleting notifications from Mongo DB
+            // runs every 60 seconds
+            TimeUnit.SECONDS.sleep(sleepTime);
+        }
+
+        @Then("^\"([^\"]*)\" has been deleted from \"([A-Za-z0-9_]+)\" database$")
```
tobiasake: Really hard for understand what is going to be deleted in test step descirption nor the code. You can adde something general description before regex, e.g. SubscriptionName "^\"([^\"]*)\"", then it would be more clearer what is happening here.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,225 @@
...
+        public void has_been_deleted_from_database(String collection, String database) {
```
tobiasake: Same comment regarding the function name.

tobiasake: What are you checking here, if you are checking a specific Document in MongoDb,, then it is more properly done by querying the specific Document instead of checking if Database and collection is empty.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,225 @@
...
+            LOGGER.debug("Checking " + collection + " in " + database);
+            List<String> result = mongoDBHandler.getAllDocuments(database, collection);
+            assertEquals("Database is not empty.", true, result.isEmpty());
+        }
+
+        // SCENARIO @TestNotificationRetries
+
+        @Given("^Subscription is created$")
+        public void create_subscription_object() throws IOException, JSONException {
+            LOGGER.debug("Starting scenario @TestNotificationRetries.");
```
tobiasake: What EI Logic are you testing here? Looks like you are testing your own code only.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,225 @@
...
+                    String.valueOf(clientAndServer.getPort()));
+
+            subscriptionObject = new ObjectMapper().readTree(subscriptionStr);
+            assertEquals(false, subscriptionObject.get("notificationMeta").toString().contains("{port}"));
+        }
+
+        @When("^I fail to inform subscriber$")
+        public void trigger_notification() throws IOException {
+            JsonNode aggregatedObject = getJSONFromFile(AGGREGATED_OBJECT_FILE_PATH);
+            informSubscription.informSubscriber(aggregatedObject.toString(), subscriptionObject);
```
tobiasake: This test step will always succeed?

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
```diff
@@ -0,0 +1,225 @@
...
+                    .withPath(ENDPOINT), Format.JSON);
+            JSONArray requests = new JSONArray(retrievedRequests);
+
+            // received requests include number of retries
+            assertEquals(2, requests.length());
+        }
+
+        @Then("^Missed notification is in database$")
+        public void missed_notification_is_in_database() {
+            List<String> result = mongoDBHandler.getAllDocuments(
```
tobiasake: If you know the subscriptionName or some other parameters from the object, its better to query on that parameter when checking the MongoDb. Then the check is more properly done.

src/functionaltests/resources/AggregatedObject.json
```diff
@@ -0,0 +1,34 @@
+{
```
tobiasake: I think you have these files under "src/test/resources/..." . I think its better to reuse files instead of duplicate files. Or have you done and any specific changes in the files for your tests?
This comment applies to the other json files also.

src/functionaltests/java/com/ericsson/ei/notifications/ttl/TestTTLSteps.java
e-pettersson-ericsson: Hmm, this test step is general because a check is done to verify both aggregated object and missed notification has been deleted, that's why I couldn't find a general noun to fit both in the  method name. 
An option is to directly spell out missed notification and aggregated object into the test steps and check both databases at the same time and not have regex in feature.

src/main/java/com/ericsson/ei/subscriptionhandler/SpringRestTemplate.java
```diff
@@ -73,13 +75,19 @@ public int postDataMultiValue(String notificationMeta, MultiValueMap<String, Str
                         headers);
                 response = rest.postForEntity(notificationMeta, request, JsonNode.class);
             }
+        } catch (HttpClientErrorException e) {
+            LOGGER.error("HTTP-request failed, bad request!\n When trying to connect to URL: " + notificationMeta);
+            return HttpStatus.BAD_REQUEST.value();            
+        } catch (HttpServerErrorException e) {
+            LOGGER.error("HTTP-request failed, internal server error!\n When trying to connect to URL: " + notificationMeta);
```
tobiasake: Include e.getMessage() in Logger.error message. This apples for both these two catches.

PR #140, opened 2018-08-01
SEPARATE: informSubscription method

ned29:


src/main/java/com/ericsson/ei/subscriptionhandler/InformSubscription.java
```diff
@@ -102,25 +102,8 @@ public void informSubscriber(String aggregatedObject, JsonNode subscriptionJson)
...
-
-                } else {
-
-                    mapNotificationMessage.add(objNode.get("formkey").toString().replaceAll(REGEX, ""), jmespath
-                            .runRuleOnEvent(objNode.get("formvalue").toString().replaceAll(REGEX, ""), aggregatedObject)
-                            .toString().replaceAll(REGEX, ""));
-                }
-            }
-        }
+        MultiValueMap<String, String> mapNotificationMessage = mapNotificationMessage(aggregatedObject, subscriptionJson);
```
evasiba-ericsson: reset key and value fields to empty fields at the end of this method and add a comment that we add this because we do not want to leak authentication details.

src/main/java/com/ericsson/ei/subscriptionhandler/InformSubscription.java
```diff
@@ -58,9 +56,11 @@
 @Component
 public class InformSubscription {
 
-    private static final Logger LOGGER = (Logger) LoggerFactory.getLogger(InformSubscription.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(InformSubscription.class);
     // Regular expression for replacement unexpected character like \"|
     private static final String REGEX = "^\"|\"$";
+    private static String key = "";
```
evasiba-ericsson: these fields need to be reset at the end of the method inform subscriber.

PR #144, opened 2018-08-09
Add  functional tests for reconnecting with message bus

e-pettersson-ericsson:


src/functionaltests/java/com/ericsson/ei/utils/AMQPBrokerManager.java
```diff
@@ -6,28 +6,35 @@
 import org.apache.qpid.server.BrokerOptions;
 
 public class AMQPBrokerManager {
+    public Boolean isRunning = false;
```
ned29: make it boolean

src/functionaltests/java/com/ericsson/ei/utils/TestConfigs.java
```diff
@@ -1,62 +1,53 @@
...
 import de.flapdoodle.embed.mongo.distribution.Version;
 import de.flapdoodle.embed.mongo.tests.MongodForTestsFactory;
 
 public class TestConfigs {
 
-    private static AMQPBrokerManager amqpBroker;
-    private static ConnectionFactory cf;
-    private Connection conn;
-    private MongodForTestsFactory testsFactory;
+    final static Logger LOGGER = (Logger) LoggerFactory.getLogger(TestConfigs.class);
```
ned29: make it private and delete (Logger) unneeded casting here

src/functionaltests/java/com/ericsson/ei/utils/TestConfigs.java
```diff
@@ -1,62 +1,53 @@
...
+
     private MongoClient mongoClient = null;
 
-    final static Logger LOGGER = (Logger) LoggerFactory.getLogger(TestConfigs.class);
+    protected static HashMap<Integer, AMQPBrokerManager> amqpBrokerPool = new HashMap<Integer, AMQPBrokerManager>();
```
ned29: should be  - Map < Integer, AMQPBrokerManager > amqpBrokerPool = new HashMap<>();
we should use interface realization for collections and diamonds type in new HashMap<>

src/main/java/com/ericsson/ei/rmqhandler/RmqHandler.java
```diff
@@ -208,8 +208,8 @@ public void confirm(CorrelationData correlationData, boolean ack, String cause)
     }
 
     public void publishObjectToWaitlistQueue(String message) {
-        log.info("publishing message to message bus...");
-        rabbitMqTemplate().convertAndSend(message);
+        log.info("Publishing message to message bus...");
```
ned29: set level to debug

src/main/java/com/ericsson/ei/rmqhandler/RmqHandler.java
```diff
@@ -218,8 +218,9 @@ public void close() {
             container.destroy();
             factory.destroy();
         } catch (Exception e) {
-            log.info("exception occured while closing connections");
+            log.info("Exception occured while closing connections");
```
ned29: the same

src/functionaltests/java/com/ericsson/ei/rabbitmq/RabbitMQTestConnectionSteps.java
```diff
@@ -0,0 +1,92 @@
...
+    public void message_bus_is_restarted() throws Exception {
+        amqpBroker.startBroker();
+    }
+
+    @Then("^I can send events which are put in the waitlist$")
+    public void can_send_events_which_are_put_in_the_waitlist() throws Exception {
+        LOGGER.debug("Sending eiffel events");
+        int waitListSize = 0;
+        List<String> eventNames = getEventNamesToSend();
+        long maxTime = System.currentTimeMillis() + TimeUnit.SECONDS.toMillis(30000);
```
evasiba-ericsson: you are converting 30000 seconds to milliseconds. Is it what you want?

change to 30.

src/functionaltests/java/com/ericsson/ei/utils/TestContextInitializer.java
```diff
@@ -12,12 +21,12 @@
 
     @Override
     public void initialize(ConfigurableApplicationContext ac) {
-        try {
-            amqpBroker();
-            mongoClient();
 
+        try {
+            createAmqpBroker(); // get instance from pool where? here, or from tests steps?
```
evasiba-ericsson: remove comment

src/main/java/com/ericsson/ei/rmqhandler/RmqHandler.java
```diff
@@ -117,6 +112,7 @@
     private SimpleMessageListenerContainer container;
     private SimpleMessageListenerContainer waitlistContainer;
 
+
     @Bean
     ConnectionFactory connectionFactory() {
         com.rabbitmq.client.ConnectionFactory connectionFactory = new com.rabbitmq.client.ConnectionFactory();
```
evasiba-ericsson: Please remove the package from the type. Nicer to have it in imports. I do not know how we managed to have it here.

src/functionaltests/java/com/ericsson/ei/utils/TestConfigs.java
```diff
@@ -1,62 +1,53 @@
...
-        cf.setPort(port);
-        cf.setHandshakeTimeout(600000);
-        cf.setConnectionTimeout(600000);
-        conn = cf.newConnection();
+        AMQPBrokerManager amqpBroker = new AMQPBrokerManager(qpidConfig.getAbsolutePath(), Integer.toString(port));
+
         LOGGER.debug("Started embedded message bus for tests on port: " + port);
+        amqpBroker.startBroker();
+
+        amqpBrokerPool.put(port, amqpBroker); // add new amqp broker to pool
```
evasiba-ericsson: comments always before code

src/functionaltests/java/com/ericsson/ei/utils/TestConfigs.java
```diff
@@ -75,33 +66,12 @@ public void setAuthorization() {
...
-            // the connection is closed we just receive the
-            // exception and go on
-        }
 
-        if (mongoClient != null)
-            mongoClient.close();
-        if (testsFactory != null)
-            testsFactory.shutdown();
+    public static AMQPBrokerManager getBrokerFromPool(int port) {
+        return amqpBrokerPool.get(port);
```
tobiasake: How should this pool work? One test class execution per isolated MessageBus or several tests can share same messagebus.
If you want one to have one isolated messagebus per test, then some sort of locking mechanism need to be implemented around this, I guess.

src/main/java/com/ericsson/ei/rmqhandler/RmqHandler.java
e-pettersson-ericsson: It seems this is a workaround for using org.springframework.amqp.rabbit.connection.CachingConnectionFactory and com.rabbitmq.client.ConnectionFactory together as they are otherwise incompatible.

src/functionaltests/java/com/ericsson/ei/utils/TestConfigs.java
evasiba-ericsson: It is not the concern of the test. The initializer fixes it for you. The application just need to use the port from the system property.

src/main/java/com/ericsson/ei/rmqhandler/RmqHandler.java
evasiba-ericsson: no it is not.

This star import 

import org.springframework.amqp.rabbit.connection.*;

needs to be replaced with:

import org.springframework.amqp.rabbit.connection.CachingConnectionFactory;
import com.rabbitmq.client.ConnectionFactory;

src/functionaltests/java/com/ericsson/ei/utils/TestConfigs.java
tobiasake: Then why having AMQP broker in a hasmap when its initialized from the test and shutdown from the test? There is no meaning of having this in hash map when it only used by one test. Normally a pool of services works according:
1. Test Frameworks start and starts 20(defined by setting) services and add these services to the pool(pool has locking mechanism for every services).
2. 25 parallel tests starts and 20 of these tests occupies and hold locks for each service in pool. 5 remaining tests acquire lock on service from pool, but pool has no free service and these test is put in wait queue for acquiring lock on a service.
3. When a test is finished, test release the lock for the service and one test from the lock queue can now acquire lock on the newly lock released service.
4. And it continues in the same way for all tests that is the lock queue.
This my view of how a pool normally works.
I guess the idea was to reduce the test execution time by reducing start/boot up time of message bus, and reusing already running message buses.

src/main/java/com/ericsson/ei/rmqhandler/RmqHandler.java
e-pettersson-ericsson: The implementation as it is now requires both com.rabbitmq.client.ConnectionFactory  and org.springframework.amqp.rabbit.connection.ConnectionFactory. It is not enough to change the imports as different connectionfactories are used in the code. An option is to rewrite the whole implementation but that should maybe me moved to another PR. I've changed the naming of the different packages used to hopefully clarify which is what in the code.

src/functionaltests/java/com/ericsson/ei/utils/AMQPBrokerManager.java
```diff
@@ -6,28 +6,35 @@
...
+        isRunning = true;
     }
 
     public void stopBroker() {
         broker.shutdown();
+        isRunning = false;
+    }
+
+    public String getPort() {
+        return this.port;
```
ned29: we have lombok lib for getters

src/functionaltests/java/com/ericsson/ei/utils/TestConfigs.java
```diff
@@ -1,62 +1,54 @@
...
-    private static AMQPBrokerManager amqpBroker;
-    private static ConnectionFactory cf;
-    private Connection conn;
-    private MongodForTestsFactory testsFactory;
+    private final static Logger LOGGER = LoggerFactory.getLogger(TestConfigs.class);
+
     private MongoClient mongoClient = null;
 
-    final static Logger LOGGER = (Logger) LoggerFactory.getLogger(TestConfigs.class);
+    protected static HashMap<Integer, AMQPBrokerManager> amqpBrokerMap = new HashMap<>();
```
ned29: should be - Map< Integer, AMQPBrokerManager > amqpBrokerMap = new HashMap<>();

src/functionaltests/java/com/ericsson/ei/rabbitmq/RabbitMQTestConnectionSteps.java
```diff
@@ -0,0 +1,90 @@
...
+            eventManager.sendEiffelEvents(EIFFEL_EVENTS, eventNames);
+            TimeUnit.SECONDS.sleep(2);
+            waitListSize = dbManager.waitListSize();
+        }
+        assertEquals(4, waitListSize);
+    }
+
+    @AfterClass
+    public void tearDown() throws IOException {
+        LOGGER.debug("Shutting down AMQP broker after tests");
```
evasiba-ericsson: Are you sure this is run.

src/functionaltests/java/com/ericsson/ei/utils/BaseRunner.java
```diff
@@ -0,0 +1,12 @@
+package com.ericsson.ei.utils;
+
+import org.junit.AfterClass;
+
+public class BaseRunner {
+
+    @AfterClass()
+    public static void shutdownAmqpBroker() {
+        String rabbitMQPort = System.getProperty("rabbitmq.port");
+        TestContextInitializer.removeBroker(Integer.parseInt(rabbitMQPort));
```
evasiba-ericsson: do not do the parsing here. Create in initializer a new method removeBroker that take a string and do parsing there before calling the other method removeBroker that takes the int.

PR #145, opened 2018-08-10
Improve authentication functional tests

Christoffer-Cortes:


src/functionaltests/java/com/ericsson/ei/subscriptions/authentication/AuthenticationSteps.java
```diff
@@ -2,104 +2,110 @@
...
+
+    @LocalServerPort
+    private int applicationPort;
+    private HttpRequest httpRequest;
+    private ResponseEntity<String> response;
+
+    @Before("@RESTWithCredentials,@RESTWithSession")
+    public void beforeScenario() throws Throwable {
+        httpRequest = new HttpGetRequest();
+        httpRequest.setUrl("http://localhost:").setPort(applicationPort).setEndpoint("/auth/logout");
```
tobiasake: I don't remember if we decided to reuse functional tests in Systemtest level.
If its going to be used in system test level,, hostname need to be fetched from configuration.
On Dockerhost/Kubernetes each component will have different hostname alias.

src/functionaltests/java/com/ericsson/ei/subscriptions/authentication/TestAuthenticationRunner.java
```diff
@@ -7,9 +7,7 @@
 @RunWith(Cucumber.class)
 @CucumberOptions(features = "src/functionaltests/resources/features/authentication.feature", glue = {
     "com.ericsson.ei.subscriptions.authentication" }, plugin = { "pretty",
-    "html:target/cucumber-reports/TestSubscriptionCRUDRunner" }, monochrome = false)
+    "html:target/cucumber-reports/TestAuthenticationRunner" }, monochrome = false)
```
e-pettersson-ericsson: Remove monochrome = false, as it is false by default

src/functionaltests/java/com/ericsson/ei/subscriptions/authentication/AuthenticationSteps.java
tobiasake: This comment applies for all hardcoded localhosts occurrences.

src/functionaltests/java/com/ericsson/ei/subscriptions/repeatHandler/TestSubscriptionRepeatHandlerRunner.java
```diff
@@ -7,7 +7,7 @@
 @RunWith(Cucumber.class)
 @CucumberOptions(features = "src/functionaltests/resources/features/subscriptionRepeatHandler.feature", glue = {
         "com.ericsson.ei.subscriptions.repeatHandler" }, plugin = { "pretty",
-        "html:target/cucumber-reports/TestSubscriptionRepeatHandleRunner" }, monochrome = false)
+        "html:target/cucumber-reports/TestSubscriptionRepeatHandlerRunner" }, monochrome = false)
```
e-pettersson-ericsson: Same comment as previous.

src/functionaltests/java/com/ericsson/ei/utils/HttpDeleteRequest.java
```diff
@@ -1,34 +1,14 @@
...
 
     /**
      * Build together a httpDelete object
-     * */
-    public ResponseEntity<String>  build() {
-        HttpDelete httpDelete = new HttpDelete(url + port + endpoint);
-
-        return restApi.getResponse(httpDelete);
+     */
+    public ResponseEntity<String> build() {
```
tobiasake: Change functional name on build() function to something more valid name, since actual also do the request. Maybe change to name "performRequest()" or something. Same comment for all HTTP Request classes.
Update the java doc so it describes more what it do.

src/functionaltests/java/com/ericsson/ei/utils/HttpGetRequest.java
```diff
@@ -1,31 +1,14 @@
...
-    @Getter @Setter private String endpoint;
-    private SubscriptionRestAPI restApi = new SubscriptionRestAPI();
-
+public class HttpGetRequest extends HttpRequest {
 
     /**
      * Build together a httpGet object
-     * */
+     */
     public ResponseEntity<String> build() {
```
tobiasake: same comment here.

src/functionaltests/java/com/ericsson/ei/utils/HttpPostRequest.java
```diff
@@ -1,70 +1,14 @@
...
-        params = new StringEntity(jsonParams.toString(), "UTF-8");
-        return this;
-    }
+public class HttpPostRequest extends HttpRequest {
 
     /**
      * Build together a httpPost object
-     * */
+     */
     public ResponseEntity<String> build() {
```
tobiasake: same comment here.

src/functionaltests/java/com/ericsson/ei/utils/HttpDeleteRequest.java
evasiba-ericsson: build is necessary when implementing build pattern. However do request should not be done here. It should however be a new story to refactor these classes.

src/functionaltests/java/com/ericsson/ei/subscriptions/authentication/AuthenticationSteps.java
evasiba-ericsson: right now we do not know how the configuration will be passed but it is good to implement a method that returns the host and then we can modify that method later to fetch host from property

src/functionaltests/java/com/ericsson/ei/utils/HttpDeleteRequest.java
Christoffer-Cortes: I didn't see your comment until now. I remade the way HttpRequest works without split classes. The thing is that when I started to split up the test methods I noticed that they weren't modular. I needed to be able to add a credentials header for example to GET or POST request without knowing which method it is. That is why I made the parent class HttpRequest and then I began to see how much they had in common and that the only thing that split them appart was one line. Should i revert back and find another solution?

src/functionaltests/java/com/ericsson/ei/utils/SubscriptionRestAPI.java
```diff
@@ -16,9 +16,21 @@
 
 public class SubscriptionRestAPI {
```
evasiba-ericsson: since you anyway doing refactoring change the name of this class to HttpExecutor or other more appropriate name since this has nothing to do with subscriptions

src/functionaltests/java/com/ericsson/ei/utils/SubscriptionRestAPI.java
```diff
@@ -31,12 +43,14 @@
         String jsonContent = "";
```
evasiba-ericsson: processRequest or executeRequest is a more appropriate name.

PR #159, opened 2018-08-16
Replaced MvcMock with HttpRequest and some other changes

Christoffer-Cortes:


src/functionaltests/java/com/ericsson/ei/subscriptions/crud/SubscriptionCRUDSteps.java
```diff
@@ -34,87 +32,105 @@
...
+    public void my_GET_request_with_subscription_name_at_REST_API_returns_empty_String(String name, String endPoint)
+            throws Throwable {
+        httpRequest = new HttpRequest(HttpMethod.GET);
+        httpRequest.setHost(hostName).setPort(applicationPort).setEndpoint(endPoint + name)
+                .addHeader("Accept", "application/json");
+        response = httpRequest.performRequest();
+        JsonNode node = eventManager.getJSONFromString(response.getBody());
+        int size = node.get("foundSubscriptions").size();
+        String notFound = node.get("notFoundSubscriptions").get(0).asText();
+        assertEquals(0, size);
```
tobiasake: Since you are only comparing numbers, it will be hard to simple see what it failed on. Add a message regarding what failed to the assertEquals: e.g. assertEquals(<message>, expected, actual)

src/functionaltests/java/com/ericsson/ei/subscriptions/crud/SubscriptionCRUDSteps.java
```diff
@@ -34,87 +32,105 @@
...
+                .setBody(jsonArray.toString());
+        response = httpRequest.performRequest();
     }
 
-    @Then("^I get response code of (\\d+)$")
-    public void i_get_response_code_of(int statusCode) throws Throwable {
-        Assert.assertEquals(statusCode, result.getResponse().getStatus());
+    @Then("^I get response code of (\\d+)")
+    public void i_get_response_code_of(int statusCode) {
+        assertEquals(HttpStatus.valueOf(statusCode), response.getStatusCode());
```
tobiasake: You can add failed message to the assertEquals(<message>, expected, actual)

Christoffer-Cortes: I can add a message but I'm not sure how to make it any clearer than the actual assert failure. The test is also set with pretty which makes it very clear at what scenario and step it failed.

tobiasake: ok. If its clear what failed, then keep it as it is. I have seen assertions earlier which comapares to integers that it result is actual "0" is not "5",, but it was not clear what failed. My opinion is that test result should be as informative as possible so a user that don't that much about the test case can understand the failure in 1 or 2 seconds by just reading the failure message and result.

Christoffer-Cortes: In the number comparison I can add what failed if the assert fails but on the status code failure it's clear what was expected and at what scenario/step it failed.

PR #160, opened 2018-08-17
Fix rule check endpoint

e-edling-ericsson:
- Changed query parameters to json in body.
- Fixed tests affected by this.


src/main/java/com/ericsson/ei/controller/RuleCheckControllerImpl.java
```diff
@@ -72,32 +73,35 @@
...
     @CrossOrigin
     @ApiOperation(value = "To execute rule on JSON", response = String.class)
     public ResponseEntity<?> updateRulesRuleCheck(
-            @ApiParam(value = "JMESPath rule", required = true) @RequestParam String rule,
-            @ApiParam(value = "JSON object", required = true) @RequestBody String jsonContent) {
+            @ApiParam(value = "JSON object", required = true) @RequestBody RuleCheckBody body) {
+        JSONObject rule = new JSONObject(body.getRule().getAdditionalProperties());
+        JSONObject event = new JSONObject(body.getEvent().getAdditionalProperties());
+
+        String ruleString = rule.toString().replaceAll("\"", "");
```
tobiasake: Doesn't these JSON objects has a function that is called "asText()", I think JsonNode object has this function anyway , which removes these extra quotations. That function could be used, if I now understand this situation correct.

e-edling-ericsson: The reason for the replacement is that jmespath.compile(rule) that is executed is very sensitive and doesn't allow quotations in the expression. asText() sometimes makes empty texts and i don't really know why.

PR #162, opened 2018-08-21
Change parameter to body in free style query

eroohen:


src/main/java/com/ericsson/ei/controller/model/ListEventsJson.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({
+
+})
+public class ListEventsJson {
+
```
tobiasake: All these classes seems to have same member and functions,, can't this be inherit from an extended class?

src/main/java/com/ericsson/ei/controller/model/Options.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({
+
+})
+public class Options {
+
```
tobiasake: same comment here.

eroohen: These are auto-generated classes based on the raml schema. Changes I do here will be overwritten in 1 sec.

PR #164, opened 2018-09-07
DURACI-11027 free style query requires implementation knowledge

sericcson:


src/main/resources/application.properties
```diff
@@ -53,6 +53,10 @@ spring.data.mongodb.database: eiffel_intelligence
 server.session-timeout: 1200
 sessions.collection.name: sessions
 
+# name of the aggregated object use by user
```
evasiba-ericsson: change the name of the property to search.query.prefix.

Explain that this prefix needs to prefix to each criteria in order for the search to work.

PR #165, opened 2018-09-11
Change: subscription template

ned29:


src/main/resources/templates/subscriptionsTemplate.json
```diff
@@ -1,5 +1,5 @@
 [{
-    "subscriptionName" : "Subscription1",
+    "subscriptionName" : "Subscription1__Only_numbers_letters_and_underscore_are_allowed",
```
tobiasake: what is this used for? for test? If test, then it should be placed under src/test/resources.

ned29: Regarding my story, Sergio requested include to template which digital are allowed, please look into my story

ned29: Should I do it or it will do someone else?
Also i think i should close this pr.

evasiba-ericsson: This should be as a comment in the subscription on WIKI

PR #191, opened 2018-11-15
Customiseable mail-notification subject

saif-ericsson:
### Description of the Change
Today all mail notifications use the same subject. This change allow that user can implement the subject per subscription in subscription JSON object.
### Benefits

Users can have notification subject of their choice

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:


src/main/java/com/ericsson/ei/subscriptionhandler/SendMail.java
```diff
@@ -107,6 +111,10 @@ public void sendMail(String receiver, String mapNotificationMessage) throws Mess
         return emailAdd;
     }
 
+    public String getSubject(String subject, String emailSubject) {
```
Christoffer-Cortes: String subject already exists in the scope of the class as an instance variable, does it really need to be explicitly sent to this method?

src/main/java/com/ericsson/ei/subscriptionhandler/InformSubscription.java
```diff
@@ -105,6 +105,8 @@ public void informSubscriber(String aggregatedObject, JsonNode subscriptionJson)
 		String notificationType = getSubscriptionField("notificationType", subscriptionJson);
 		String notificationMeta = getSubscriptionField("notificationMeta", subscriptionJson);
 
+		String subject = getSubscriptionField("emailSubject", subscriptionJson);
```
a-breid-ericsson: As a user I would want a default subject to be added in case I did not add one. Discuss what would be best. Could be "Subscription triggered: <Subscription name>"

src/main/java/com/ericsson/ei/subscriptionhandler/SendMail.java
```diff
@@ -107,6 +111,10 @@ public void sendMail(String receiver, String mapNotificationMessage) throws Mess
         return emailAdd;
     }
 
+    public String getSubject(String emailSubject) {
```
a-breid-ericsson: What do this function do?

saif-ericsson: Well, this function return default subject if one is not provided in the subscription.

src/main/java/com/ericsson/ei/subscriptionhandler/InformSubscription.java
saif-ericsson: Currently it is "Email Subscription Notification"... not bad

src/main/java/com/ericsson/ei/subscriptionhandler/SendMail.java
e-edling-ericsson: One liners are cool but it's pretty hard to read, even when you know how it works. I would suggest writing it out instead

PR #213, opened 2018-12-20
Better response when invalid user deletes wrong subscription

a-breid-ericsson:
Fixing better response when a user tries to delete someone elses subscription. Maybe the gui should handle this before the request are made, but anyhow right now the message is better.

### Applicable Issues
A response was displayed for a user that the subscription did not exist if a user tried to delete a subscription made by someone else. But id does exist. Or at leas I can see it, maybe it is just in my mind...

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:


src/main/java/com/ericsson/ei/controller/SubscriptionControllerImpl.java
```diff
@@ -56,6 +57,7 @@
 
     private static final String SUBSCRIPTION_NOT_FOUND = "Subscription is not found";
     private static final String SUBSCRIPTION_ALREADY_EXISTS = "Subscription already exists";
+    private static final String INVALID_USER = "Invalid user! You must be logged in as the creater to a subscription modify it.";
```
e-pettersson-ericsson: creater -> creator

e-pettersson-ericsson: Also, I would prefer if we add "Unauthorized" instead of invalid user.

Christoffer-Cortes: 'You must be logged in as the creator of a subscription to modify it'

a-breid-ericsson: Done.

a-breid-ericsson: Done.

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -133,19 +134,25 @@ public boolean modifySubscription(Subscription subscription, String subscription
...
 
     @Override
-    public boolean deleteSubscription(String subscriptionName) {
-        String ldapUserName = (authenticate) ? HttpSessionConfig.getCurrentUser() : "";
-        String query = generateQuery(subscriptionName, ldapUserName);
-        boolean result = subscriptionRepository.deleteSubscription(query);
+    public boolean deleteSubscription(String subscriptionName) throws NotOwnerException {
+        String ldapUserName = (ldapEnabled) ? HttpSessionConfig.getCurrentUser() : "";
+        String deleteQuery = generateQuery(subscriptionName, ldapUserName);
+        String fetchQuery = generateQuery(subscriptionName, "");
```
tobiasake: fetchquery seems not to be used, or?

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -57,7 +58,7 @@
     private String repeatFlagHandlerCollection;
 
     @Value("${ldap.enabled}")
-    private boolean authenticate;
+    private boolean ldapEnabled;
```
tobiasake: if you change name here, then you can change the name in all other places where the "authenticate" name is used. So we use same name in all places.

src/main/java/com/ericsson/ei/controller/SubscriptionControllerImpl.java
```diff
@@ -170,11 +172,17 @@
...
+
+            try {
+                if (subscriptionService.deleteSubscription(subscriptionName)) {
+                    LOG.debug("Subscription is deleted successfully: " + subscriptionName);
+                } else {
+                    LOG.error("Subscription to delete is not found: " + subscriptionName);
+                    errorMap.put(subscriptionName, SUBSCRIPTION_NOT_FOUND);
+                }
+            } catch (NotOwnerException e) {
+                LOG.error("Unauthorized, wrong username: " + subscriptionName);
```
evasiba-ericsson: you log two error messages for the same exception. I thin it is better to remove the error logging here and keep the error logging where the exception is created.

Alternatively create an exception with the message you log when you create it and log it here.

src/main/java/com/ericsson/ei/services/SubscriptionService.java
a-breid-ericsson: Fixed.

src/main/java/com/ericsson/ei/services/SubscriptionService.java
a-breid-ericsson: Done

src/main/java/com/ericsson/ei/controller/SubscriptionControllerImpl.java
a-breid-ericsson: Done.

PR #247, opened 2019-02-11
Stepped component versions in docker-compose files.

tobiasake:
Stepping component versions to latest versions in docker-compose files.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
 before_script:
   - source docker-env.bash
   - docker-compose up -d
-  - sleep 20
+  - sleep 60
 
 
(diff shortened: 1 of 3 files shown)


tobiasake:
> Why do we have two docker-compose? We need to maintain both.

According to the review comments we wanted it simple for a user to start up a slim EI environment for the user local build docker image/container. User can of course configure it self. The other docker-compose file is for integrations and system test and includes fake Mail Smtp server etc. That docker-compose file + docker-env.bash should be placed under src/integrationstest/docker folder.

PR #261, opened 2019-03-12
User can perform CRUD on anonymous subscriptions

saif-ericsson:
We have had some issues that resulted in subscriptions saved in backend without user.
The problem is that everyone should be able to delete or edit these subscriptions but no one is allowed.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:


src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -207,4 +220,27 @@ private String generateQuery(String subscriptionName, String ldapUserName) {
...
+
+    /**
+     * This method finds whether a given subscription has an owner
+     *
+     * @param subscriptionName-
+     *            subscription name
+     * @return a boolean
+     * @throws SubscriptionNotFoundException
+     */
+
```
a-breid-ericsson: There should not be a line between javadock and function

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -138,20 +146,25 @@ public boolean modifySubscription(Subscription subscription, String subscription
     @Override
     public boolean deleteSubscription(String subscriptionName) throws AccessException {
         String ldapUserName = (ldapEnabled) ? HttpSessionConfig.getCurrentUser() : "";
-        String deleteQuery = generateQuery(subscriptionName, ldapUserName);
+        String deleteQuery;
+        boolean ownerExist = doSubscriptionOwnerExist(subscriptionName);
+        if (ownerExist) {
```
a-breid-ericsson: This if-statement is duplicated, I would prefere if the if-statement was in the generateQuery function.

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -207,4 +220,27 @@ private String generateQuery(String subscriptionName, String ldapUserName) {
         }
```
a-breid-ericsson: In this functiuion you may edit the if-statement like this.

boolean ownerExist = doSubscriptionOwnerExist(subscriptionName);
if (!ldapUserName.isEmpty() && ownerExist) {
}

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -207,4 +220,27 @@ private String generateQuery(String subscriptionName, String ldapUserName) {
...
+
+    private boolean doSubscriptionOwnerExist(String subscriptionName) {
+        boolean ownerExist = true;
+        try {
+            if (getSubscription(subscriptionName).getLdapUserName()
+                                                 .isEmpty()) {
+                ownerExist = false;
+            }
+        } catch (SubscriptionNotFoundException e) {
+            LOG.error(e.getMessage());
```
e-edling-ericsson: If an exception occurs, should ownerexist still be set to true? How can an owner exist when subscription could not be found?

src/main/java/com/ericsson/ei/services/SubscriptionService.java
saif-ericsson: done

src/main/java/com/ericsson/ei/services/SubscriptionService.java
saif-ericsson: I rephrased it

src/main/java/com/ericsson/ei/services/SubscriptionService.java
saif-ericsson: fixed

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -197,14 +204,34 @@ private boolean cleanSubscriptionRepeatFlagHandlerDb(String subscriptionNameQuer
      *            name of the current user
      * @return a String object
      */
-
     private String generateQuery(String subscriptionName, String ldapUserName) {
         String query = String.format(SUBSCRIPTION_NAME, subscriptionName);
-        if (!ldapUserName.isEmpty()) {
+        if (!ldapUserName.isEmpty() && ldapUserName != null) {
```
tobiasake: Shouldn't this if statement be reversed?, since "ldaUseName" can be null when isEmpty() is executed.
Should be reversed to:
"if (ldapUserName != null && !ldapUserName.isEmpty())"

saif-ericsson: You are right... Null check should be first

Fixed

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -197,14 +204,34 @@ private boolean cleanSubscriptionRepeatFlagHandlerDb(String subscriptionNameQuer
...
+     *
+     * @param subscriptionName-
+     *            subscription name
+     * @return a boolean
+     * @throws SubscriptionNotFoundException
+     */
+    private boolean doSubscriptionOwnerExist(String subscriptionName) {
+        boolean ownerExist = false;
+        try {
+            if (!getSubscription(subscriptionName).getLdapUserName()
```
a-breid-ericsson: I believe this if statement should be in one line, we do not use Christoffers formatter any more since it had few flaws. Use Standard Eiffel formatter.

src/main/java/com/ericsson/ei/services/SubscriptionService.java
```diff
@@ -113,17 +113,22 @@ public boolean doSubscriptionExist(String subscriptionName) {
     public boolean modifySubscription(Subscription subscription, String subscriptionName) {
         ObjectMapper mapper = new ObjectMapper();
         Document result = null;
+        String query;
         try {
             String stringSubscription = mapper.writeValueAsString(subscription);
             String ldapUserName = (ldapEnabled) ? HttpSessionConfig.getCurrentUser() : "";
-            String query = generateQuery(subscriptionName, ldapUserName);
+            boolean ownerExist = doSubscriptionOwnerExist(subscriptionName);
+            if (!ownerExist) {
```
a-breid-ericsson: This if-statement is duplicated, please move the if-statement to be inside the generateQuery function, cause in that case then it only needs to be added once.

src/main/java/com/ericsson/ei/services/SubscriptionService.java
saif-ericsson: Anyway, I changed it as a one liner...

src/main/java/com/ericsson/ei/services/SubscriptionService.java
saif-ericsson: I have created a new method so that IF is used only once

PR #276, opened 2019-05-23
regEx changed in backend to match those in frontend

saif-ericsson:
### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
+++ b/src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
@@ -76,10 +76,15 @@ public class SubscriptionValidator {
      * @param subscriptionName
      */
     private static void validateSubscriptionName(String subscriptionName) throws SubscriptionValidationException {
-        String regex = "^[A-Za-z0-9_]+$";
+        // When this regExp need to be changed then remember to change the one in the
+        // back-end (invalidSubscriptionNameRegex in subscription.js), which do the same
+        // invalid subscription name check. The two
+        // regEx always need to be the same for ensuring the same check.
+        // /(\W)/ Is a regEx that matches anything that is not [A-Z,a-z,0-8] and _.
+        String invalidSubscriptionNameRegex = "(\\W)";
         if (subscriptionName == null) {
             throw new SubscriptionValidationException("Required field SubscriptionName has not been set");
-        } else if (!Pattern.matches(regex, subscriptionName)) {
+        } else if (Pattern.matches(invalidSubscriptionNameRegex, subscriptionName)) {
             throw new SubscriptionValidationException("Wrong format of SubscriptionName: " + subscriptionName);
         }
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -76,10 +76,15 @@ public static void validateSubscription(Subscription subscription) throws Subscr
      * @param subscriptionName
      */
     private static void validateSubscriptionName(String subscriptionName) throws SubscriptionValidationException {
-        String regex = "^[A-Za-z0-9_]+$";
+        // When this regExp need to be changed then remember to change the one in the
```
a-breid-ericsson: Maybe we should discuss in the team how we would like to handle regexp's. Maybe it should be added in the Eiffel Commons repository so the same regexp's may be loaded from there, Maybe even have them in a property file like emailValidationRegex and then any Eiffel products may use the same regexp.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -180,7 +185,12 @@ private static void RestPostMediaType(String restPostMediaType) throws Subscript
      * @param email
      */
     public static void validateEmail(String email) throws SubscriptionValidationException {
-        final Pattern VALID_EMAIL_ADDRESS_REGEX = Pattern.compile("^[A-Z0-9._%+-]+@[A-Z0-9.-]+\\.[A-Z]{2,6}$",
+        // When this regExp need to be changed then remember to change the one in the
+        // back-end (validEmailRegExpression in subscription.js), which do the same
+        // email validation check. The two
+        // regEx always need to be the same for ensuring the same check.
+        String validEmailRegExpression = "^(([^<>()\\[\\]\\\\.,;:\\s@\"]+(\\.[^<>()\\[\\]\\\\.,;:\\s@\"]+)*)|(\".+\"))@((\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}])|(([a-zA-Z\\-0-9]+\\.)+[a-zA-Z]{2,}))$";
```
tobiasake: This is a very very long regex! Have you unit tests for all cases this regex should match and fail on?
Is this supposed to regex on both list of mails (comma separated) and a single mail.
I think the code becomes a lot more readable if you have regexp for a single mail and them split the mail-list on comma and validate one mail at time.
Now not human understandable regexp for a normal developer.
What is valid characters in mail: '.', '-', '@' , '[a-zA-Z0-9]',,, am I missing something?

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
tobiasake: Eiffel-Common is good place the validation code.
Regexp source should be mapped/defined in som Json Map file.
It would be good if Frontend/bridge has a rest entrypoint so JavaScript code in web-browser can use same Regexp for validations in Javascript code.

saif-ericsson: Yes, I agree we need a discussion.... As per bug requirement, it is now same in both backend and frontend. So I intend to create a story after the discussion.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
saif-ericsson: Well, unit tests are already there, I just took the regex(the longer one) from frontend and put in the backend. 
In fact, it is for single email validation. But, I agree it is difficult to comprehend. We will discuss it before taking next step.

e-edling-ericsson: Of course this is a long string, this is a string created by communities for valid email addresses. We shouldn't make them ourselves.

saif-ericsson:
> Looks ok, make some manual tests with different mail addresses, both valid and invalid. And create stories for the solution we discussed earlier about putting regex centrally in EiffelCommon/EI-Backend.


Two stories are created and regEx tested

PR #277, opened 2019-05-28
Regex from common to validate subscription in EI back-end

saif-ericsson:
We need to validate subscription in both front-end and back-end. Currently, we are using regular expressions which are part of code in both front-end and back-end. Thus, if we need to change regular expressions, we need to change at both places. It is difficult to maintain. Therefore, it is decided that we create a library (for regex) at eiffel-commons and both front-end and back-end will extract it from there.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:


pom.xml
```diff
@@ -68,9 +68,9 @@
 
     <dependencies>
         <dependency>
-            <groupId>com.github.eiffel-community</groupId>
-            <artifactId>eiffel-commons</artifactId>
-            <version>0.0.4</version>
+            <groupId>com.github.ericsson</groupId>
+            <artifactId>eiffel-commons-java</artifactId>
```
a-breid-ericsson: Wrong artifact groupId and artifactId

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -70,18 +75,22 @@ public static void validateSubscription(Subscription subscription) throws Subscr
...
-        // back-end (invalidSubscriptionNameRegex in subscription.js), which do the same
-        // invalid subscription name check. The two
-        // regEx always need to be the same for ensuring the same check.
-        // /(\W)/ Is a regEx that matches anything that is not [A-Z,a-z,0-8] and _.
-        String invalidSubscriptionNameRegex = "(\\W)";
+        // Here the regEx used is :/(\W)/ that matches anything that is not
+        // [A-Z,a-z,0-8] and _.
+        String invalidSubscriptionNameRegex = null;
+        try {
+            invalidSubscriptionNameRegex = (String) RegExProvider.getRegExs().get("invalidName");
```
a-breid-ericsson: You may get it as string directly and there is no need to cats it. 
RegExProvider.getRegExs().getString("invalidName");
Should do the work for you.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -70,18 +75,22 @@ public static void validateSubscription(Subscription subscription) throws Subscr
...
     private static void validateSubscriptionName(String subscriptionName) throws SubscriptionValidationException {
-        // When this regExp need to be changed then remember to change the one in the
-        // back-end (invalidSubscriptionNameRegex in subscription.js), which do the same
-        // invalid subscription name check. The two
-        // regEx always need to be the same for ensuring the same check.
-        // /(\W)/ Is a regEx that matches anything that is not [A-Z,a-z,0-8] and _.
-        String invalidSubscriptionNameRegex = "(\\W)";
+        // Here the regEx used is :/(\W)/ that matches anything that is not
+        // [A-Z,a-z,0-8] and _.
+        String invalidSubscriptionNameRegex = null;
```
a-breid-ericsson: What happens if the next try catch fails and this is null in the validation?

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -179,19 +190,19 @@ private static void RestPostMediaType(String restPostMediaType) throws Subscript
...
-        // When this regExp need to be changed then remember to change the one in the
-        // back-end (validEmailRegExpression in subscription.js), which do the same
-        // email validation check. The two
-        // regEx always need to be the same for ensuring the same check.
-        String validEmailRegExpression = "^(([^<>()\\[\\]\\\\.,;:\\s@\"]+(\\.[^<>()\\[\\]\\\\.,;:\\s@\"]+)*)|(\".+\"))@((\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}])|(([a-zA-Z\\-0-9]+\\.)+[a-zA-Z]{2,}))$";
-        final Pattern VALID_EMAIL_ADDRESS_REGEX = Pattern.compile(validEmailRegExpression,
-                Pattern.CASE_INSENSITIVE);
+        String validEmailRegEx = null;
+        try {
+            validEmailRegEx = (String) RegExProvider.getRegExs().get("validEmail");
```
a-breid-ericsson: See above comment about casting.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -179,19 +190,19 @@ private static void RestPostMediaType(String restPostMediaType) throws Subscript
...
+        } catch (JSONException | IOException e) {
+            LOGGER.error("Error message: " + e.getMessage(), e);
+        }
+        final Pattern VALID_EMAIL_ADDRESS_REGEX = Pattern.compile(validEmailRegEx, Pattern.CASE_INSENSITIVE);
```
a-breid-ericsson: see above comment if regex is null.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
Christoffer-Cortes: In the documentation it says that if no mapping is found a NullPointerException is thrown. So you can either catch and log that or you can return a default value like this "getString("invalidName","") and check that it is not empty in an else if. You still need to log something if it can't validate though so perhaps catching the null pointer is best.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
tobiasake: I thought the regex object in EiffelCommon was going to have same Key values as in Subscription object as in the Json object of regex. Which means an iterator can iterate the Subscription Json object and use that key to retrieve the corresponding regex from regex json object and do the regex. This make the code working no matter how many fields it is in Subscription json object or in Regex Object. If regex is added and/or if subscription fields added, no code changes need to be done, only Subscription Json file and Regex Json file need to updated with the new regex/subscription fields.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
saif-ericsson: It is handled now

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -179,24 +192,29 @@ private static void RestPostMediaType(String restPostMediaType) throws Subscript
...
-                            + email + "]");
+        String validEmailRegEx = null;
+        try {
+            validEmailRegEx = RegExProvider.getRegExs().getString("validEmail");
+        } catch (JSONException | IOException e) {
+            LOGGER.error("Error message: " + e.getMessage(), e);
+        }
+
+        if (validEmailRegEx == null || validEmailRegEx.isEmpty()) {
+            throw new SubscriptionValidationException("A valid regEx for email is not provided");
```
evasiba-ericsson: Change message to :
A valid regular expression for subscription email validation is not provided

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -70,29 +75,35 @@ public static void validateSubscription(Subscription subscription) throws Subscr
...
+        try {
+            invalidSubscriptionNameRegex = RegExProvider.getRegExs().getString("invalidName");
+        } catch (JSONException | IOException e) {
+            LOGGER.error("Error message: " + e.getMessage(), e);
+        }
+
         if (subscriptionName == null) {
             throw new SubscriptionValidationException("Required field SubscriptionName has not been set");
+        } else if (invalidSubscriptionNameRegex == null || invalidSubscriptionNameRegex.isEmpty()) {
+            throw new SubscriptionValidationException("A valid regEx is not provided");
```
evasiba-ericsson: change message to 
A valid regular expression for validating subscription name is not provided.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -70,29 +75,36 @@ public static void validateSubscription(Subscription subscription) throws Subscr
...
+        try {
+            invalidSubscriptionNameRegex = RegExProvider.getRegExs().getString("invalidName");
+        } catch (JSONException | IOException e) {
+            LOGGER.error("Error message: " + e.getMessage(), e);
+        }
+
         if (subscriptionName == null) {
             throw new SubscriptionValidationException("Required field SubscriptionName has not been set");
+        } else if (invalidSubscriptionNameRegex == null || invalidSubscriptionNameRegex.isEmpty()) {
+            throw new SubscriptionValidationException(
```
a-breid-ericsson: Move this regex check to it´s own function that may be called for both name validation and email validation, thus there is not need to have duplicated code.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -179,24 +193,30 @@ private static void RestPostMediaType(String restPostMediaType) throws Subscript
...
-        Matcher matcher = VALID_EMAIL_ADDRESS_REGEX.matcher(email);
-        if (!(matcher.matches())) {
+        String validEmailRegEx = null;
+        try {
+            validEmailRegEx = RegExProvider.getRegExs().getString("validEmail");
+        } catch (JSONException | IOException e) {
+            LOGGER.error("Error message: " + e.getMessage(), e);
+        }
+
+        if (validEmailRegEx == null || validEmailRegEx.isEmpty()) {
```
a-breid-ericsson: See previous comment.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -179,24 +193,30 @@ private static void RestPostMediaType(String restPostMediaType) throws Subscript
...
             throw new SubscriptionValidationException(
-                    "Notification type is set to [MAIL] but the given notificatioMeta contains an invalid e-mail ["
-                            + email + "]");
+                    "A valid regular expression for subscription email validation is not provided");
+        } else {
```
a-breid-ericsson: This else is not needed, since if above if-statement is true then it will raise an exception and not run the code below.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
saif-ericsson: done

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
saif-ericsson: done

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
saif-ericsson: done

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
saif-ericsson: But we are taking only two Subscription object values (Subscription name and email)to validate... So these values need to be called by name for retrieving regexs from common... or ...?

Christoffer-Cortes: I think he means calling the regex keys subscriptionName and notificationMeta instead of calling the m invalidName and validEmail. Though notificationMeta can be either a URL or email...

tobiasake: I develop from a future perspective,,, so maintenance becomes as simple and fast as it get.
Your propose require more code to maintain and more code to change  when in future more regex is added or subscriptions fields added. Since you stated that meta field has several use case and different regex/checks, it is ok to have it like you have it. But change the field/regex key names according to what we decided.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -70,29 +76,32 @@ public static void validateSubscription(Subscription subscription) throws Subscr
...
-        // When this regExp need to be changed then remember to change the one in the
-        // back-end (invalidSubscriptionNameRegex in subscription.js), which do the same
-        // invalid subscription name check. The two
-        // regEx always need to be the same for ensuring the same check.
-        // /(\W)/ Is a regEx that matches anything that is not [A-Z,a-z,0-8] and _.
-        String invalidSubscriptionNameRegex = "(\\W)";
+        // Here the regEx used is :/(\W)/ that matches anything that is not
+        // [A-Z,a-z,0-8] and _.
+        String invalidSubscriptionNameRegex = null;
+        invalidSubscriptionNameRegex = RegExProvider.SUBSCRIPTION_NAME;
```
e-pettersson-ericsson: According to the comment above the regex does not match a subscription name, but everything else right? Then I find is misleading that the regex name for it says "SUBSCRIPTION_NAME" implying that the regex contains the opposite ... 

Either change the name of the constant or, what I prefer, change the regex to contain only allowed values and then update this check.

saif-ericsson: Well, this specific regex was used (I think) to spill out the wrong characters, especially in the front-end subscription form, to inform user. The name, I agree it doesnt give the clear picture, is used to match the one used in the subscription object. So if other agree to use different name then I will change it.

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -179,19 +190,20 @@ private static void RestPostMediaType(String restPostMediaType) throws Subscript
...
-        final Pattern VALID_EMAIL_ADDRESS_REGEX = Pattern.compile(validEmailRegExpression,
-                Pattern.CASE_INSENSITIVE);
+        String validEmailRegEx = null;
+        validEmailRegEx = RegExProvider.NOTIFICATION_META;
+
+        if (validEmailRegEx == null || validEmailRegEx.isEmpty()) {
+            throw new SubscriptionValidationException(
+                    "A valid regular expression for subscription email validation is not provided");
+        }
+        final Pattern VALID_EMAIL_ADDRESS_REGEX = Pattern.compile(validEmailRegEx, Pattern.CASE_INSENSITIVE);
```
tobiasake: Shouldn't local variable be with lower letters?

src/main/java/com/ericsson/ei/subscription/SubscriptionValidator.java
```diff
@@ -70,29 +76,32 @@ public static void validateSubscription(Subscription subscription) throws Subscr
...
      */
     private static void validateSubscriptionName(String subscriptionName) throws SubscriptionValidationException {
-        // When this regExp need to be changed then remember to change the one in the
-        // back-end (invalidSubscriptionNameRegex in subscription.js), which do the same
-        // invalid subscription name check. The two
-        // regEx always need to be the same for ensuring the same check.
-        // /(\W)/ Is a regEx that matches anything that is not [A-Z,a-z,0-8] and _.
-        String invalidSubscriptionNameRegex = "(\\W)";
+        // Here the regEx used is :/(\W)/ that matches anything that is not
+        // [A-Z,a-z,0-8] and _.
```
tobiasake: Why keeping these comments? Since the regex is in Common library. These comment will not be synced with common library changes.

PR #316, opened 2019-08-12
Fix ERQueryService URI build bug

### LINKED ISSUE #299: ER query service URL path issue
### Description

When adding the er.url property it is very sensitive how you specify the URL. If there is a "/" in the end of the URL it will result in an incorrect URL when the ER query service is building the entire URL since it is using "+" operator when building the entire URL. It will need to be changed to use a URL builder so that it doesnt matter if you end it with "/" or not.
### Motivation

Otherwise it will be unclear whether the operator of the system should the the URL with "/" or not.
### Exemplification

It should work with both following er.url values:
`http://<address>:<port>/search`
and
`http://<address>:<port>/search/`
### Benefits

Easier to configure EI with less risk for mistakes.

e-edling-ericsson:
Needs https://github.com/eiffel-community/eiffel-commons/pull/24 merged and released first.

### Applicable Issues

closes #299
### Description of the Change

- Fix bug where the URI to ERQueryService was built up incorrect. Changed the rest template to the HttpRequest in Eiffel commons which handles this as well as cleaner code.
- Fix tests to work correctly with new request library. 

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: 
Erik Edling erik.edling@ericsson.com

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,9 +72,9 @@
 
     <dependencies>
         <dependency>
-            <groupId>com.github.ericsson</groupId>
-            <artifactId>eiffel-commons-java</artifactId>
-            <version>0.0.15-SNAPSHOT</version>
+            <groupId>com.github.eiffel-community</groupId>
+            <artifactId>eiffel-commons</artifactId>
+            <version>0.0.15</version>
         </dependency>
 
         <dependency>
@@ -511,7 +511,7 @@
                     </execution>
                 </executions>
                 <configuration>
-                    <forkCount>1</forkCount>
+                    <forkCount>3</forkCount>
                     <reuseForks>false</reuseForks>
                     <skipTests>${skipTests}</skipTests>
                     <skipITs>${skipITs}</skipITs>
--- a/pom.xml
+++ b/pom.xml
@@ -72,9 +72,9 @@
 
     <dependencies>
         <dependency>
-            <groupId>com.github.eiffel-community</groupId>
-            <artifactId>eiffel-commons</artifactId>
-            <version>0.0.13</version>
+            <groupId>com.github.ericsson</groupId>
+            <artifactId>eiffel-commons-java</artifactId>
+            <version>0.0.15-SNAPSHOT</version>
         </dependency>
 
         <dependency>
@@ -511,7 +511,7 @@
                     </execution>
                 </executions>
                 <configuration>
-                    <forkCount>3</forkCount>
+                    <forkCount>1</forkCount>
                     <reuseForks>false</reuseForks>
                     <skipTests>${skipTests}</skipTests>
                     <skipITs>${skipITs}</skipITs>
--- a/src/main/java/com/ericsson/ei/erqueryservice/ERQueryService.java
+++ b/src/main/java/com/ericsson/ei/erqueryservice/ERQueryService.java
@@ -14,54 +14,49 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
-*/
+ */
 package com.ericsson.ei.erqueryservice;
 
-import com.fasterxml.jackson.databind.JsonNode;
-import lombok.Getter;
-import org.apache.commons.lang3.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Value;
-import org.springframework.boot.web.client.RestTemplateBuilder;
-import org.springframework.http.HttpEntity;
-import org.springframework.http.HttpHeaders;
-import org.springframework.http.HttpMethod;
-import org.springframework.http.MediaType;
-import org.springframework.http.ResponseEntity;
-import org.springframework.stereotype.Component;
-import org.springframework.web.client.RestClientException;
-import org.springframework.web.client.RestOperations;
-import org.springframework.web.util.UriComponents;
-import org.springframework.web.util.UriComponentsBuilder;
-
-import javax.annotation.PostConstruct;
-import java.net.URI;
+import java.net.MalformedURLException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 
+import javax.annotation.PostConstruct;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.apache.http.entity.ContentType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import com.ericsson.eiffelcommons.utils.HttpRequest;
+import com.ericsson.eiffelcommons.utils.HttpRequest.HttpMethod;
+import com.ericsson.eiffelcommons.utils.ResponseEntity;
+
+import lombok.Getter;
 /**
  * @author evasiba
  */
+
 @Component
 public class ERQueryService {
-
     private static final Logger LOGGER = LoggerFactory.getLogger(ERQueryService.class);
 
-    private RestOperations rest;
+    private HttpRequest request;
 
     @Getter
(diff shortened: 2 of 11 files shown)


pom.xml
```diff
@@ -511,7 +511,7 @@
                     </execution>
                 </executions>
                 <configuration>
-                    <forkCount>3</forkCount>
+                    <forkCount>1</forkCount>
                     <reuseForks>false</reuseForks>
                     <skipTests>${skipTests}</skipTests>
                     <skipITs>${skipITs}</skipITs>
```
e-edling-ericsson: This file should not be changed, i will revert it back once eiffel commons is merged, i will need to use 0.0.15 only.

src/main/java/com/ericsson/ei/erqueryservice/ERQueryService.java
```diff
@@ -72,21 +67,28 @@ public void setRest(RestOperations rest) {
...
-                final ResponseEntity<String> response = rest.getForEntity(erUrl, String.class, params);
+    public ResponseEntity getEventDataById(String eventId) {
+        String erUrl = null;
+
+        try {
+            if(StringUtils.isNotBlank(erBaseUrl)) {
+                request
+                    .setHttpMethod(HttpMethod.GET)
+                    .setBaseUrl(erBaseUrl)
+                    .setEndpoint("{id}")
```
tobiasake: What {id}? Shouldn't ER "events" entry point specified. Is it some hidden functionality you use that it not seen in the code?
Since method is called getEventDataById(), I guess same endpoint will be used for every call to this function, so entry point can be hardcoded into code or use a property where entrypoint is specified.

src/main/java/com/ericsson/ei/erqueryservice/ERQueryService.java
```diff
@@ -112,33 +114,33 @@ public void setRest(RestOperations rest) {
...
-                LOGGER.error("Error occurred while executing REST POST to: {} for\n{}", erUrl, requestEntity, e);
+        String uri = null;
+        try {
+            if(StringUtils.isNotBlank(erBaseUrl)) {
+                // Request Body parameters
+                final SearchParameters searchParameters = getSearchParameters(searchOption);
+                request
+                        .setHttpMethod(HttpMethod.POST)
+                        .setBaseUrl(erBaseUrl)
+                        .setEndpoint(eventId)
```
tobiasake: isn't "events" entry point missing here?,, if I remember correct eventid value follows after the "events" entrypoint.

src/main/java/com/ericsson/ei/erqueryservice/ERQueryService.java
e-edling-ericsson: It is not something i did come up with. If you look in the old code it looked like this.

`final String erUrl = URI.create(url.trim() + "/" + "{id}").normalize().toString();`

I just added it like it was but translated to the new lib.

src/main/java/com/ericsson/ei/erqueryservice/ERQueryService.java
e-edling-ericsson: I'm using the exactly same url as before, which works.

This is the old builup.
`final String erUrl = URI.create(url.trim() + "/" + eventId).normalize().toString();`

Christoffer-Cortes: The search endpoint is part of the er.url

src/main/java/com/ericsson/ei/erqueryservice/ERQueryService.java
tobiasake: Ok,, code looks very confusing. Let it be as it is then.
But the endpoint should be "/events/{eventId}"
Looking at this code now, first half of endpoint ,"/events", seems to be included in erBaseUrl.
Which is wrong and confusing, if you ask me.
This can be fixed in another PR then.

src/main/java/com/ericsson/ei/erqueryservice/ERQueryService.java
tobiasake: Ok,, maybe same comment as previous comment.

src/main/java/com/ericsson/ei/erqueryservice/SearchParameters.java
```diff
@@ -60,8 +65,41 @@ public void setUlt(final List<LinkType> ult) {
...
     }
 
+    /**
+     * Returns the search parameters as a json string
+     * @return String
+     * @throws IOException
+     */
+    public String getAsJsonString() throws IOException {
+        ArrayList<String> dltStringArray = convertSearchParametersToArrayList(dlt);
+        ArrayList<String> ultStringArray = convertSearchParametersToArrayList(ult);
```
tobiasake: what is "dlt" amd "ult"? in the StringArray names.
Use more understanding names.

src/test/java/com/ericsson/ei/erqueryservice/test/ERQueryServiceTest.java
```diff
@@ -81,83 +77,36 @@
...
-        erQueryService.setRest(rest);
-        searchOption = SearchOption.UP_STREAM;
-        given(rest.exchange(Mockito.any(URI.class), Mockito.any(HttpMethod.class), Mockito.any(HttpEntity.class),
-                Mockito.any(Class.class))).willAnswer(
-                        returnRestExchange(Mockito.any(URI.class), Mockito.any(HttpMethod.class),
-                                Mockito.any(HttpEntity.class), Mockito.any(Class.class)));
-        ResponseEntity<JsonNode> result = erQueryService.getEventStreamDataById(eventId, searchOption, limitParam,
-                levels, isTree);
+    public void testErQueryUpstream() throws ClientProtocolException, URISyntaxException, IOException {
+//        when(httpExecutor.executeRequest(any(HttpRequestBase.class))).thenAnswer(validateRequest(Mockito.any(HttpRequestBase.class)));
```
Christoffer-Cortes: Remove comment?

src/main/java/com/ericsson/ei/erqueryservice/ERQueryService.java
e-edling-ericsson: Ye i agree with that. In order to change that, we will need to make changes in the configuration so that the endpoint is not part of er.url.

src/main/java/com/ericsson/ei/erqueryservice/SearchParameters.java
e-edling-ericsson: I would love to change the names, but their names in the request are called dlt and ult and i actually don't know what they stand for ❓

Christoffer-Cortes: Something upstream/downstream. =)

Christoffer-Cortes: Emelie said that it's probably downstreamLinkType and she's is right. They are declared as List<LinkType> dlt, so it seems likely. But I don't know if that name is any more descriptive to the use and purpose of the variable.

e-edling-ericsson: So the question is, is it nicer to use downstreamLinkType /upstreamLinkType or keep it as dlt/ult? I'm split in my opinion 😄 Should i keep going for dlt/ult since the REST parameters are called that way(which makes them map together) or downstreamLinkType /upstreamLinkType since they are more descriptive (but might be harder to relate to the REST parameters dlt/ult?)

tobiasake: dlt and ult is bad names,, write good names or comment about what dlt and ult is. I guess you pick the best and correct way of doing this :) ,, since we have discussed this kind of things lately :)

tobiasake: even if the the restpost parameter is dlt.. for making request line shorter.

e-edling-ericsson: Done

PR #357, opened 2019-09-25
Move RabbitMQ properties to a separate class

henning-roos:
### Description of the Change
RabbitMQ related properties now located in separate class RMQProperties.
Removed unnecessary tests that were duplicated.

### Benefits
Separation of concern: RMQProperties extracted from RMQHandler logic.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: @Christoffer-Cortes @e-edling-ericsson @e-pettersson-ericsson @henning-roos @a-breid-ericsson @m-linner-ericsson @saif-ericsson @tobiasake


src/main/java/com/ericsson/ei/controller/model/ParseInstanceInfoEI.java
```diff
@@ -19,6 +19,7 @@
 import com.ericsson.ei.erqueryservice.ERQueryService;
 import com.ericsson.ei.handlers.ObjectHandler;
 import com.ericsson.ei.handlers.RMQHandler;
```
m-linner-ericsson: I guess we don't need the import for RMQHandler now?

tobiasake:
Can't we create a class that collects all Spring/Java properties,, so we get only one object api for all properties. This was an idea long time to implement, but has always ended with "that we can do later".

e-edling-ericsson:
@tobiasake I don't quite agree that we should have one class for every property out there. Maybe one package though were all classes that reads properties exists. Otherwise it could be quite hard to know which properties the application requires without looking through every possible class in the project.

The reason I don't want them in the same big class is that it will be huge. I think it's good to divide it up "logically" into it's own property classes.

tobiasake:
Well, you can structure properties class similar as it structured in application.properties file. I guess its ok to have in different classes in same package. But if a class need properties from different properties objects,, then it will be alot of Autowiring, which can be unnecessary from some users point of view or opinions, which prefer all properties was collected in same properties class and only one autowire properties object is required to access all properties.

src/main/java/com/ericsson/ei/controller/model/ParseInstanceInfoEI.java
henning-roos: Done in upcoming commit

PR #396, opened 2019-11-25
Fix bug not validating Subscription name.

a-breid-ericsson:
### Applicable Issues
Pattern.matcher was used to try to validate subscription name, but this matches whole name and does not check parts of the name for invalid characters.

### Description of the Change
Subscription name validation now correctly matches invalid characters.
Improved tests that did not catch this bug.
Added more invalid names to test against.
Added more invalid emails to test against.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:


src/test/java/com/ericsson/ei/subscription/SubscriptionValidatorTest.java
```diff
@@ -71,64 +72,52 @@ public SubscriptionValidatorTest() {
...
-    }
-
-    @Test
-    public void validateSubscriptionNameInvalidNameTest() throws Exception {
-        String subscriptionName = "Kal--l[[e1";
-        try {
-            invokeMethod(SubscriptionValidator.class, "validateSubscriptionName", subscriptionName);
-        } catch (SubscriptionValidationException e) {
-            return;
+        String[] validSubscriptionNameList = { "Kalle", "kalle", "KALLE", "kall_e", "Kalle1",
```
m-linner-ericsson: Here and elsewhere, is there a reason not to use `@Parameterized` ([example](https://www.tutorialspoint.com/junit/junit_parameterized_test.htm)) and not catch the exception

src/test/java/com/ericsson/ei/subscription/SubscriptionValidatorTest.java
```diff
@@ -142,8 +131,7 @@ public void validateNotificationMessageValidMessageTest() throws Exception {
             invokeMethod(SubscriptionValidator.class, "validateNotificationMessageKeyValues",
                     subscription.getNotificationMessageKeyValues(), "");
         } catch (SubscriptionValidationException e) {
-            assertTrue(e.getMessage(), false);
-            return;
+            fail(e.getMessage());
```
m-linner-ericsson: Here and elsewhere, why bother to catch?

src/test/java/com/ericsson/ei/subscription/SubscriptionValidatorTest.java
```diff
@@ -247,11 +235,12 @@ public void validateNotificationMessageInvalidMessage3Test() throws Exception {
         subscription.getNotificationMessageKeyValues().add(notificationMessageKeyValue);
         try {
             invokeMethod(SubscriptionValidator.class, "validateNotificationMessageKeyValues",
-                    subscription.getNotificationMessageKeyValues(), MediaType.APPLICATION_FORM_URLENCODED.toString());
+                    subscription.getNotificationMessageKeyValues(),
+                    MediaType.APPLICATION_FORM_URLENCODED.toString());
+            fail("Expected subscriptionvalidation to throw exception.");
```
m-linner-ericsson: Here and elsewhere why not use `@Test(expected = SubscriptionValidationException.class)`

src/test/java/com/ericsson/ei/subscription/SubscriptionValidatorTest.java
a-breid-ericsson: It is complicated and need to be run with another runner than we use for this test.

src/test/java/com/ericsson/ei/subscription/SubscriptionValidatorTest.java
a-breid-ericsson: Maybe not needed since any exception will fail to test, but updating old tests is not prioritized.

src/test/java/com/ericsson/ei/subscription/SubscriptionValidatorTest.java
a-breid-ericsson: Maybe since expected exceptions may be tested, but updating old tests is not prioritized.

src/test/java/com/ericsson/ei/subscription/SubscriptionValidatorTest.java
m-linner-ericsson: Isn't it easier to remove the catch then to add the fail?

src/test/java/com/ericsson/ei/subscription/SubscriptionValidatorTest.java
m-linner-ericsson: The question should then be: Why add a fail when you can use the test expection

src/test/java/com/ericsson/ei/subscription/SubscriptionValidatorTest.java
m-linner-ericsson: I was only wondering as you have added the for loop. My view is not to add your own logic if there is something available.

PR #410, opened 2019-12-10
Improvement of MongoDB object cleanup.

tobiasake:
### Applicable Issues

We have got some reports of MongoDb connection issues while MongoDb connection goes down and comes up again.
Description of the Change

### Description of the Change
This change might improve the mongodb object to re-establish MongoDb connection after connection issues.

### Possible Drawbacks
This will produce more logs and exceptions during tests and might make Travis-CI to abort test execution due to much logs is printed to console/logs.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/src/main/java/com/ericsson/ei/handlers/MongoDBHandler.java
+++ b/src/main/java/com/ericsson/ei/handlers/MongoDBHandler.java
@@ -318,7 +318,6 @@ public class MongoDBHandler {
         if (mongoClient != null) {
             mongoClient.close();
         }
-        mongoClient = null;
     }
     
     /**


src/main/java/com/ericsson/ei/handlers/MongoDBHandler.java
```diff
@@ -318,7 +318,6 @@ private void closeMongoDbConecction() {
         if (mongoClient != null) {
             mongoClient.close();
         }
-        mongoClient = null;
```
m-linner-ericsson: I don't like this solution especially when there are no tests to prove it. But otherwise it feels wrong

Christoffer-Cortes: It is closed in the if statement above this. Is that not enough?

tobiasake: Yes, but as we discussed, this MongoDb will be removed/rewritten in Anders pull request in future release.

m-linner-ericsson: I don't like to change code without tests. So I am still against it

tobiasake: I have tested it locally, by taking down and up the MongoDb instance manaully and MongoDb connection comes up.
Tested both before this change and after the change.
Since this code will be removed, I will not spend time on writing test for it.
Then someone else can write test or we skip this change and close the PR.

a-breid-ericsson: I will not fix this issue in the PR that is up now, it is already large, but I may take the bug story and implement a more solid solution.

m-linner-ericsson: I would vote to close the PR then

PR #429, opened 2020-01-27
Added default sender and subject values for email notification.

navya108:
### Applicable Issues
https://github.com/eiffel-community/eiffel-intelligence/issues/401

### Description of the Change
Added sender and subject default values for email notification.

### Benefits
We will always have email sender and subject defined, even if users do not defined them in their application.properties.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
+++ b/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
@@ -195,16 +195,10 @@ public class SubscriptionNotificationSteps extends FunctionalTestBase {
         eventManager.sendEiffelEvents(EIFFEL_EVENTS_JSON_PATH, eventNamesToSend);
     }
 
-    @Then("^Then Notification email contains <our expected> values")
-    public void notification_email_contains_expected_values() {
-        sender = "noreply@ericsson.com";
-        subject = "Email Subscription Notification";
-    }
-
-    @Then("^Notification email contains default values$")
-    public void notification_email_contains_default_values() {
-        assertEquals("noreply@ericsson.com", sender);
-        assertEquals("Email Subscription Notification", subject);
+    @Then("^Then Notification email contains {String} and {String} values")
+    public void notification_email_contains_expected_values(String sender, String subject) {
+        assertEquals(sender, emailSender.getSender());
+        assertEquals(subject, emailSender.getSubject());
     }
 
     @Then("^Mail subscriptions were triggered$")
--- a/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
+++ b/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
@@ -70,6 +70,9 @@ public class SubscriptionNotificationSteps extends FunctionalTestBase {
     @Value("${email.sender}")
     private String sender;
 
+    @Value("${email.subject}")
+    private String subject;
+
     @Value("${aggregations.collection.name}")
     private String aggregatedCollectionName;
 
@@ -187,6 +190,18 @@ public class SubscriptionNotificationSteps extends FunctionalTestBase {
         eventManager.sendEiffelEvents(EIFFEL_EVENTS_JSON_PATH, eventNamesToSend);
     }
 
+    @Then("^Default values mail notification are assigned$")
+    public void default_values_mail_notification() {
+        if (sender == null && subject == null) {
+            sender = "noreply@ericsson.com";
+            subject = "Email Subscription Notification";
+        } else if (sender == null && subject != null) {
+            sender = "noreply@ericsson.com";
+        } else if (sender != null && subject == null) {
+            subject = "Email Subscription Notification";
+        }
+    }
+
     @Then("^Mail subscriptions were triggered$")
     public void mail_subscriptions_were_triggered() {
         LOGGER.debug("Verifying received emails.");
@@ -196,6 +211,8 @@ public class SubscriptionNotificationSteps extends FunctionalTestBase {
         for (SmtpMessage email : emails) {
             // assert correct sender.
             assertEquals("Assert correct email sender: ", email.getHeaderValue("From"), sender);
+            // assert correct subject.
+            assertEquals("Assert correct email subject: ", email.getHeaderValue("Subject"), subject);
             // assert given test case exist in body.
             assert (email.getBody().contains("TC5"));
         }
--- a/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
+++ b/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
@@ -17,6 +17,7 @@ import java.util.concurrent.TimeUnit;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.junit.Ignore;
+import org.mockito.InjectMocks;
 import org.mockserver.client.MockServerClient;
 import org.mockserver.integration.ClientAndServer;
 import org.mockserver.model.Format;
@@ -34,6 +35,7 @@ import com.dumbster.smtp.SimpleSmtpServer;
 import com.dumbster.smtp.SmtpMessage;
 import com.ericsson.ei.mongo.MongoCondition;
 import com.ericsson.ei.mongo.MongoDBHandler;
+import com.ericsson.ei.notifications.EmailSender;
 import com.ericsson.ei.utils.FunctionalTestBase;
 import com.ericsson.ei.utils.HttpRequest;
 import com.fasterxml.jackson.databind.JsonNode;
@@ -90,6 +92,9 @@ public class SubscriptionNotificationSteps extends FunctionalTestBase {
 
     @Autowired
     private MongoDBHandler mongoDBHandler;
+
+    @InjectMocks
+    private EmailSender emailSender;
 
     private SimpleSmtpServer smtpServer;
     private ClientAndServer restServer;
@@ -190,15 +195,11 @@ public class SubscriptionNotificationSteps extends FunctionalTestBase {
         eventManager.sendEiffelEvents(EIFFEL_EVENTS_JSON_PATH, eventNamesToSend);
     }
 
-    @Then("^Default values mail notification are assigned$")
+    @Then("^Default values mail notification are added$")
     public void default_values_mail_notification() {
-        if (sender == null && subject == null) {
-            sender = "noreply@ericsson.com";
-            subject = "Email Subscription Notification";
-        } else if (sender == null && subject != null) {
-            sender = "noreply@ericsson.com";
-        } else if (sender != null && subject == null) {
-            subject = "Email Subscription Notification";
+        if (emailSender.getSender() != null && emailSender.getSubject() != null) {
+            assertEquals("noreply@ericsson.com", emailSender.getSender());
+            assertEquals("Email Subscription Notification", emailSender.getSubject());
         }
     }
 
--- a/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
+++ b/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
@@ -197,9 +197,15 @@ public class SubscriptionNotificationSteps extends FunctionalTestBase {
 
     @Then("^Default values mail notification are added$")
     public void default_values_mail_notification() {
+        if (emailSender.getSender() != null && emailSender.getSubject() != null) {
             assertEquals("noreply@ericsson.com", emailSender.getSender());
             assertEquals("Email Subscription Notification", emailSender.getSubject());
+        } else if (emailSender.getSender() != null && emailSender.getSubject() == null) {
+            assertEquals("noreply@ericsson.com", emailSender.getSender());
+        } else if (emailSender.getSender() == null && emailSender.getSubject() != null) {
+            assertEquals("Email Subscription Notification", emailSender.getSubject());
         }
+    }
 
     @Then("^Mail subscriptions were triggered$")
     public void mail_subscriptions_were_triggered() {
--- a/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
+++ b/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
@@ -195,7 +195,7 @@ public class SubscriptionNotificationSteps extends FunctionalTestBase {
         eventManager.sendEiffelEvents(EIFFEL_EVENTS_JSON_PATH, eventNamesToSend);
     }
 
-    @Then("^Then Notification email contains {String} and {String} values")
+    @Then("^Then Notification email contains ('(.*?)') and ('(.*?)') values")
     public void notification_email_contains_expected_values(String sender, String subject) {
         assertEquals(sender, emailSender.getSender());
         assertEquals(subject, emailSender.getSubject());
--- a/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
+++ b/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
@@ -197,11 +197,9 @@ public class SubscriptionNotificationSteps extends FunctionalTestBase {
 
     @Then("^Default values mail notification are added$")
     public void default_values_mail_notification() {
-        if (emailSender.getSender() != null && emailSender.getSubject() != null) {
             assertEquals("noreply@ericsson.com", emailSender.getSender());
             assertEquals("Email Subscription Notification", emailSender.getSubject());
         }
-    }
 
     @Then("^Mail subscriptions were triggered$")
     public void mail_subscriptions_were_triggered() {
--- a/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
+++ b/src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
@@ -195,16 +195,16 @@ public class SubscriptionNotificationSteps extends FunctionalTestBase {
         eventManager.sendEiffelEvents(EIFFEL_EVENTS_JSON_PATH, eventNamesToSend);
     }
 
-    @Then("^Default values mail notification are added$")
-    public void default_values_mail_notification() {
-        if (emailSender.getSender() != null && emailSender.getSubject() != null) {
-            assertEquals("noreply@ericsson.com", emailSender.getSender());
-            assertEquals("Email Subscription Notification", emailSender.getSubject());
-        } else if (emailSender.getSender() != null && emailSender.getSubject() == null) {
-            assertEquals("noreply@ericsson.com", emailSender.getSender());
-        } else if (emailSender.getSender() == null && emailSender.getSubject() != null) {
-            assertEquals("Email Subscription Notification", emailSender.getSubject());
-        }
+    @Then("^Then Notification email contains <our expected> values")
+    public void notification_email_contains_expected_values() {
+        sender = "noreply@ericsson.com";
+        subject = "Email Subscription Notification";
+    }
+
+    @Then("^Notification email contains default values$")
+    public void notification_email_contains_default_values() {
+        assertEquals("noreply@ericsson.com", sender);
+        assertEquals("Email Subscription Notification", subject);
     }
 
     @Then("^Mail subscriptions were triggered$")
--- a/src/functionaltests/resources/features/subscriptionNotification.feature
+++ b/src/functionaltests/resources/features/subscriptionNotification.feature
@@ -31,6 +31,6 @@ Feature: Test Subscription Trigger
     And Subscriptions are created
     When I send Eiffel events
     And Wait for EI to aggregate objects
-    Then Notification email contains "noreply@ericsson.com" and "Email Subscription Notification" values
+    Then Notification email contains "noreply@domain.com" and "Email Subscription Notification" values
     Then Mail subscriptions were triggered
     And Rest subscriptions were triggered
--- a/src/functionaltests/resources/features/subscriptionNotification.feature
+++ b/src/functionaltests/resources/features/subscriptionNotification.feature
@@ -30,6 +30,6 @@ Feature: Test Subscription Trigger
     And Subscriptions are created
     When I send Eiffel events
     And Wait for EI to aggregate objects
-    Then Default values mail notification are assigned
+    Then Default values mail notification are added
     Then Mail subscriptions were triggered
     And Rest subscriptions were triggered
--- a/src/functionaltests/resources/features/subscriptionNotification.feature
+++ b/src/functionaltests/resources/features/subscriptionNotification.feature
@@ -8,6 +8,7 @@ Feature: Test Subscription Trigger
     And Subscriptions are created
     When I send Eiffel events
     And Wait for EI to aggregate objects
+    Then Notification email contains <our expected> values
     Then Mail subscriptions were triggered
     And Rest subscriptions were triggered
     When I send one previous event again
@@ -30,6 +31,6 @@ Feature: Test Subscription Trigger
     And Subscriptions are created
     When I send Eiffel events
     And Wait for EI to aggregate objects
-    Then Default values mail notification are added
+    Then Notification email contains default values
     Then Mail subscriptions were triggered
     And Rest subscriptions were triggered
--- a/src/functionaltests/resources/features/subscriptionNotification.feature
+++ b/src/functionaltests/resources/features/subscriptionNotification.feature
@@ -31,6 +31,6 @@ Feature: Test Subscription Trigger
     And Subscriptions are created
     When I send Eiffel events
     And Wait for EI to aggregate objects
-    Then Notification email contains default values
+    Then Notification email contains "noreply@ericsson.com" and "Email Subscription Notification" values
     Then Mail subscriptions were triggered
     And Rest subscriptions were triggered
--- a/src/functionaltests/resources/features/subscriptionNotification.feature
+++ b/src/functionaltests/resources/features/subscriptionNotification.feature
@@ -22,3 +22,14 @@ Feature: Test Subscription Trigger
     When I send one previous event again
     And Wait for EI to aggregate objects
     And Failed notification db should contain 2 objects
+
+  @AddedDefaultValuesSenderAndSubject
+  Scenario: Test default values mail notification
+    Given The REST API is up and running
+    And Mail server is up
+    And Subscriptions are created
+    When I send Eiffel events
+    And Wait for EI to aggregate objects
+    Then Default values mail notification are assigned
+    Then Mail subscriptions were triggered
+    And Rest subscriptions were triggered
--- a/src/functionaltests/resources/features/subscriptionNotification.feature
+++ b/src/functionaltests/resources/features/subscriptionNotification.feature
@@ -31,6 +31,6 @@ Feature: Test Subscription Trigger
     And Subscriptions are created
     When I send Eiffel events
     And Wait for EI to aggregate objects
-    Then Notification email contains "noreply@domain.com" and "Email Subscription Notification" values
+    Then Notification email contains 'noreply@domain.com' and 'Email Subscription Notification' values
     Then Mail subscriptions were triggered
     And Rest subscriptions were triggered
--- a/src/main/java/com/ericsson/ei/notifications/EmailSender.java
+++ b/src/main/java/com/ericsson/ei/notifications/EmailSender.java
@@ -83,7 +83,6 @@ public class EmailSender {
      * */
     public MimeMessage prepareEmailMessage(String recipients, String mapNotificationMessage,
             String emailSubject) {
-        setDefaultValues();
         Set<String> emails = new HashSet<>();
         emails = extractEmails(recipients);
         String[] to = emails.toArray(new String[0]);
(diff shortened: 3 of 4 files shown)


src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -99,6 +99,11 @@ public MimeMessage prepareEmailMessage(String recipients, String mapNotification
      */
     private MimeMessage prepareEmail(String mapNotificationMessage, String emailSubject,
             String[] recipients) {
+        // Added default sender and subject for email notification
```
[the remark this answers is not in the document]
e-pettersson-ericsson: Perhaps extending the functionaltests for subscription notification with a new scenario to test this?

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -99,6 +99,11 @@ public MimeMessage prepareEmailMessage(String recipients, String mapNotification
...
+        if (sender.isEmpty() || subject.isEmpty()) {
+            sender = "noreply@ericsson.com";
```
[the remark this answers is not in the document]
navya108: In this case it is not possible because if we make the variable final we have to initiate it directly or in a constructor but we can't use the variables in if block.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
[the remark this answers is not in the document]
navya108: ok. I am implementing the unit and functional test cases for this functionality.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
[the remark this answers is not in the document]
e-pettersson-ericsson: Ah I meant the value only, for example: private static final String SUBJECT_VALUE = "Email notification";

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -99,6 +99,11 @@ public MimeMessage prepareEmailMessage(String recipients, String mapNotification
      */
     private MimeMessage prepareEmail(String mapNotificationMessage, String emailSubject,
             String[] recipients) {
+        // Added default sender and subject for email notification
+        if (sender.isEmpty() || subject.isEmpty()) {
```
a-breid-ericsson: This if statement is not needed since the values will never be empty, they will always have a default set at the top in the @Value field. 

Also the if statement in current state will replace the sender with default if the sender is given by the user but subject is not.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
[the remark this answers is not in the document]
navya108: Yah , If that is the case we can use the value but we can't reassign it.
In our Scenario as I am assigning the value after checking the condition so it is not possible to put as static final.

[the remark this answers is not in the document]
navya108: yah Emelie I have just tried putting static final and without if condition, its working fine. I will update the change.

src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
```diff
@@ -187,6 +190,14 @@ public void send_one_previous_event() throws Throwable {
         eventManager.sendEiffelEvents(EIFFEL_EVENTS_JSON_PATH, eventNamesToSend);
     }
 
+    @Then("^Default values mail notification are assigned$")
+    public void default_values_mail_notification() {
+        if (sender == null || subject == null) {
+            sender = "noreply@ericsson.com";
+            subject = "Email subscription notification";
+        }
+    }
```
a-breid-ericsson: Please update step "Mail subscriptions were triggered" with the method name "mail_subscriptions_were_triggered" to also validate the subject in the email, currentluy it only makes an assert on the "From" field that is the sender.

You should also test and ensure that it is possible to give only sender and not subject and the given sender should still be the value given by the user, and the other way around, a subject is given by the user but not sender.

src/functionaltests/resources/features/subscriptionNotification.feature
```diff
@@ -22,3 +22,14 @@ Feature: Test Subscription Trigger
     When I send one previous event again
     And Wait for EI to aggregate objects
     And Failed notification db should contain 2 objects
+
+  @AddedDefaultValuesSenderAndSubject
```
a-breid-ericsson: This test could be extended to test also when one of the options is given but not the other, thus the given option should not be replaced with default.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -48,12 +48,12 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(EmailSender.class);
 
     @Getter
-    @Value("${email.sender}")
-    private String sender;
+    @Value("${email.sender:noreply@ericsson.com}")
+    private static final String sender = "noreply@ericsson.com";
```
a-breid-ericsson: A "private static final" variable should always be named in uppercase. You could remove static or rename the variables.

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
```diff
@@ -43,14 +44,29 @@
...
+
     @Test(expected = NotificationFailureException.class)
     public void sendEmailThrowsException() throws Exception {
         doThrow(new MailSendException("")).when(javaMailSender).send(message);
         emailSender.sendEmail(message);
     }
 
+    @Test
+    public void testDefaultValues() {
+    if (sender == null || subject == null) {
```
a-breid-ericsson: in this case if one value is given but not the other both will be set to be default not just the missed one.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
rogpers-ericsson: Can we have ericsson.com in here?

e-pettersson-ericsson: Question is: what to have instead as a default sender ? 🤔  Any suggestions?

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
e-pettersson-ericsson: Isn't it enough to extend these unit tests with more scenarios, to test? 1) only subject was given, 2) only sender was given 3) no configuration was given - default should be added. 
Then we don't have to extend the cucumber scenarios with several scenarios, but just one 😄

src/main/java/com/ericsson/ei/notifications/EmailSender.java
rogpers-ericsson: In principle, I think this should be a configurable thing. To be strict, perhaps no default may be used?

tobiasake: I think its possible/ok to leave the sender address field empty in a mail, -> defaulting to a empty string. Only issue can be that Mail lib maybe require a non-empty string on that parameter and throw an exception. You can maybe set it to no-reply@dummy.com or something.

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
Christoffer-Cortes: Currently I don't understand what this tests. The values are fetched from the properties and they have a defined default. This means they cannot be null so the test would always fall to the else statement.

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
```diff
@@ -43,14 +44,33 @@
...
+    private String subject;
+
     @Test(expected = NotificationFailureException.class)
     public void sendEmailThrowsException() throws Exception {
         doThrow(new MailSendException("")).when(javaMailSender).send(message);
         emailSender.sendEmail(message);
     }
 
+    @Test
+    public void testDefaultValues() {
```
Christoffer-Cortes: I think this test can be removed since it currently does not seem to be testing anything.

navya108: Emelie has asked for the test case, so I have added the change.

Christoffer-Cortes: You could make an instance of EmailSender and then get SENDER and SUBJECT and make sure they are set.
Then you are testing the logic through the EmailSender class.
There is also no assertion in the test so it would accept any outcome. A test should include an assert that checks some expected outcome of the test, unless this for reason is not possible. In your case you instantiate the class, get the values and assert that they were set to the default values.
You can also check that you can also override the defaults as expected by setting the environment variables first and then make an EmailSender instance and check that SENDER and SUBJECT were indeed set to something other than default.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
e-pettersson-ericsson: Yes I wrote this issue because I noticed the lib throws an exception if the sender is empty. That is why we want a default if this is NOT configured in application.properties. Admin users will have the option to configure the sender + subject (which they probably will) but in case they forget we have default values.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -48,17 +48,31 @@
...
+    @Value("${email.subject:Email Subscription Notification}")
+    private static final String SUBJECT = "Email Subscription Notification";
 
     @Autowired
     private JavaMailSender emailSender;
 
     /**
+     * @return the sender
+     */
+    public String getSender() {
```
Christoffer-Cortes: EI uses lombok @Getter so you don't beed to explicitly define these types of methods. You can read about it here: https://www.baeldung.com/lombok-ide
You may be getting an error currently and/or be able unable to get these varibles unless installing lombok.
Once you do you won't need these methods anymore.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -48,17 +48,31 @@
...
+        return SENDER;
+    }
+
+    /**
+     * @return the subject
+     */
+    public String getSubject() {
```
Christoffer-Cortes: See above.

src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
```diff
@@ -187,6 +195,14 @@ public void send_one_previous_event() throws Throwable {
         eventManager.sendEiffelEvents(EIFFEL_EVENTS_JSON_PATH, eventNamesToSend);
     }
 
+    @Then("^Default values mail notification are added$")
+    public void default_values_mail_notification() {
+        if (emailSender.getSender() != null && emailSender.getSubject() != null) {
```
e-pettersson-ericsson: This if statement is not needed, sin we always want to check the assert. If the value is null then the assertion goes wrong either way.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
navya108: okay

src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
navya108: okay You mean this scenario is not required? shall I remove it? As I have also added the subject assert in mail_subscriptions_were_triggered method.

e-pettersson-ericsson: No the scenario should still be there, but you can remove line 200: "if (emailSender.getSender() != null && emailSender.getSubject() != null)" and always check the assert on these values.

navya108: okay

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -48,17 +48,31 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(EmailSender.class);
 
     @Getter
-    @Value("${email.sender}")
-    private String sender;
+    @Value("${email.sender:noreply@ericsson.com}")
+    private static final String SENDER = "noreply@ericsson.com";
```
a-breid-ericsson: The default value is now given 2 times, both as @Value and decleration. If Eclipse ot the IDE you are using is complaining about not declared variables then the @Value maybe will not work.

a-breid-ericsson: A test must also ensure that a given value will not be set to default.

a-breid-ericsson: After some googling I found that this will not work, a @Value from spring cannot change a final static variable, thus final static must be removed. And variable name changed to lower case :)

navya108: yes If final static is used default values cant be changed. If that is the case required  ( where irrespective of user inputs, you want sender and subject as noreply.ericsson.com and Email Subscription Notification respectively.) ...these changes are suitable.

But what I have observed is, if user gives email and subject in application.proporties . Those will not override the sender and subject values in EmailSender class as they are set to final static. So we have to check it putting ifconditions ( if email.sender is empty then assign the value ) as I have done previously of my First push.

so please let me know for which case the changes are required so that accordingly I can proceed.

Christoffer-Cortes: The variable can't be static final then. Just type them out like this:
@Value("${email.sender:noreply@ericsson.com}")
private String sender;
That should make it possible for them to fetch the values from properties.
Also note that the sender variable does not need to be explicitly set, it is already set through the @Value annotation.

a-breid-ericsson:
We had a small discussion about this issue in the team but not the entire team. We have thoughts if maybe the sollution should be to have validation of all the given input options in application.properties att application start, and in that case sender and subject must always be set (and all other required parameters in the application.properties). And if so then the issue for this will not be needed. We will discuss this further with Santosh and Raja, cause that may be a better long term sollution.

src/functionaltests/resources/features/subscriptionNotification.feature
e-pettersson-ericsson: I agree, a new sceneario in which we test that the configured properties are used and not default values ... I don't see any mention of which configuration is used in any of the features. The first scenario "SubscriptionNotificationOK" could be extended with an assert that the email notifications has the correct subject + sender (matching input configuration)

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
e-pettersson-ericsson: After seeing the funcitonal test scenario added above, I agree that this test is not really needed. This is basically testing that Spring has assigned the correct values for the properties, and we want to test our logic - not Spring's. If the annotations are used correctly (see above comments by @a-breid-ericsson and @Christoffer-Cortes ) only Spring assigns the value to the variables and we don't have to do this in the EmailSender class. But we do want to check that if no configuration was given we use the default values - and this is tested in the functional tests.

e-pettersson-ericsson:
@a-breid-ericsson, it sounds like you want to hold off completely on this PR and instead go for a new issue/PR in which we add validation of all configuration being properly set? While I do agree a proper validation of all mandatory configuration is needed, it will take time before it is merged into master, so why not have these changes in (once they are approved) while waiting for the other PR?

src/main/java/com/ericsson/ei/notifications/EmailSender.java
tobiasake: Another approach is to write a Log message that Email is disabled when mail parameter is missing or failed to connect to SMTP server. And add Status info in Frontend Status button / info box that "Email is disabled" in backend.

tobiasake: And leave the mail properties default empty and don't default anything. But when mail host property is specifued, some validation is required, e.g. senders mail address.

navya108:
As per my observation and testing, If email.sender and email.subject are empty in application.proporties. Values what we are mentioning as below 
@value("${email.sender:noreply@ericsson.com}")
private String sender;
are not getting set as default values. And still sender and subject are reflecting as Empty values.

So I have added the conditions for checking isEmpty for both subject and sender and updated the test cases.
 
( **The below case** 
@value("${email.sender:noreply@ericsson.com}")
private String sender;
is working when entire fields i.e email.sender and email.subject are not present in application.proporties.)

Please check those changes and let me know if this approach is fine or not.

src/functionaltests/resources/features/subscriptionNotification.feature
navya108: In the first scenario I have extended assert for subject even. so should I remove the new scenario?

e-pettersson-ericsson: Ah my mistake, yes I see that you extended the first scenario - good! 

Just an improvement tip: the feature for the new scenario "Test default values mail notification" there is no mention of how the configuration affects these values (i.e. first scenario we provide email configuration so EI will use those for notifications, and in next scenario we provide no configuration so EI will use default values instead) 

I imagine something like this (with the first scenario being the existing one just extended:

**Scenario 1**
Given the REST API is up and running
And email configuration is provided
We create subscriptions 
etc ...
Notification email contains \<our expected\> values

**Scenario 2**
Given the REST API is up and running
And email configuration is NOT provided
We create subscriptions 
etc ...
Notification email contains default values

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -83,13 +83,29 @@ public void sendEmail(MimeMessage message) throws NotificationFailureException {
...
         MimeMessage message = prepareEmail(mapNotificationMessage, emailSubject, to);
         return message;
     }
 
+    /*
+     * This method assigns the default values of sender and subject of email notification if values
+     * at application.proporties are empty.
+     * */
+    private void setDefaultValues() {
+        if (sender.isEmpty() && subject.isEmpty()) {
```
e-pettersson-ericsson: I think this can be simplified with only two if statements right? To first check sender and then check subject. A few less lines of code 😄

src/main/java/com/ericsson/ei/notifications/EmailSender.java
tobiasake: Remove "ericsson" from the source code. change to some thing else or skip defaulting a value.
Let the Validation Properties discussion/issue solve it later.

src/functionaltests/java/com/ericsson/ei/notifications/trigger/SubscriptionNotificationSteps.java
```diff
@@ -195,16 +195,16 @@ public void send_one_previous_event() throws Throwable {
...
-            assertEquals("noreply@ericsson.com", emailSender.getSender());
-            assertEquals("Email Subscription Notification", emailSender.getSubject());
-        } else if (emailSender.getSender() != null && emailSender.getSubject() == null) {
-            assertEquals("noreply@ericsson.com", emailSender.getSender());
-        } else if (emailSender.getSender() == null && emailSender.getSubject() != null) {
-            assertEquals("Email Subscription Notification", emailSender.getSubject());
-        }
+    @Then("^Then Notification email contains <our expected> values")
+    public void notification_email_contains_expected_values() {
+        sender = "noreply@ericsson.com";
```
e-pettersson-ericsson: This doesn't contain any assert. You can send in the two String values from the feature file and then assert that both these values are set in the notification:

assertEquals("expected subject string", emailSender.getSubject())

src/functionaltests/resources/features/subscriptionNotification.feature
```diff
@@ -8,6 +8,7 @@ Feature: Test Subscription Trigger
     And Subscriptions are created
     When I send Eiffel events
     And Wait for EI to aggregate objects
+    Then Notification email contains <our expected> values
```
e-pettersson-ericsson: Instead of \<our expected\> values I want you to explicitly write which values we are using and then they can be sent to the cucumber test from the feature file using REGEX. Check other feature files how this is done.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -96,12 +96,10 @@ public MimeMessage prepareEmailMessage(String recipients, String mapNotification
      * at application.proporties are empty.
      * */
     private void setDefaultValues() {
-        if (sender.isEmpty() && subject.isEmpty()) {
-            sender = "noreply@ericsson.com";
-            subject = "Email Subscription Notification";
-        } else if (!sender.isEmpty() && subject.isEmpty()) {
+       if (subject.isEmpty()) {
```
e-pettersson-ericsson: Much beter! I wasn't aware that Spring interpreted it as empty values when the property is declared but no value. But this approach is good.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -48,17 +48,17 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(EmailSender.class);
 
     @Getter
-    @Value("${email.sender}")
+    @Value("${email.sender:noreply@ericsson.com}")
```
tobiasake: I change "...ericsson.com" to something else, e.g. domain.com or something.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -90,6 +90,19 @@ public MimeMessage prepareEmailMessage(String recipients, String mapNotification
         return message;
     }
 
+    /*
+     * This method assigns the default values for sender and subject of email notification if values
+     * at application.proporties are empty.
+     * */
+    private void setDefaultValuesEmailSenderSubject() {
+       if (subject.isEmpty()) {
+            subject = "Email Subscription Notification";
```
tobiasake: Is this if-statements really needed since it set default to  "Email Subscription Notification" on line 55, if subject string is empty.
So whole this if-statement can be removed, if understand the code correctly.

e-pettersson-ericsson: Apparently Spring doesn't fill in default values if the property is present in application.properties but without values. It interprets "" as a value and will not override it ... So that's why this check is needed. Perhaps add this in the javadoc for the method to clarify future questions 😄

navya108: yah as per my testing I observed that even the values are given at @value in Emailsender class, those are not been assigned  when the user gives as below in application.proporties.
email.sender:
email.subject:

when I tried to print sender and subject.. still I can see those empty values.
so I have kept those if statments to handle isEmpty values for sender and subject.

This @Value("${email.sender:noreply@domain.com}") and @Value("${email.subject:Email Subscription Notification}") line is working only when the entire field email.sender and email.subject are not present in application.proporties.

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
```diff
@@ -51,9 +51,9 @@ public void sendEmailThrowsException() throws Exception {
         emailSender.sendEmail(message);
     }
 
-    @Test
+    @Test(expected = NullPointerException.class)
     public void testDefaultValues() {
-        assertEquals("noreply@ericsson.com", emailSender.getSender());
+        assertEquals("noreply@domain.com", emailSender.getSender());
         assertEquals("Email Subscription Notification", emailSender.getSubject());
```
e-pettersson-ericsson: Shouldn't this test call upon emailSender.prepareEmailMessage? Then we assert the values of subject and sender in the created message? If we send in a null value for this parameter in the test, we expect the default values to have been added. And I don't think there is a need for the expected NullPointerException then.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
e-pettersson-ericsson: I saw the javadoc was updated in a previous commit, nice!

tobiasake: Then I guess default value in @Value ":noreply@....." is not needed , since its defaulted here instead.

e-pettersson-ericsson:
I see that Codacy webhooks are not working properly, but it complains about unused imports in your PR :) I'm looking into how to fix that Codacy checks are also enabled for PRs.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -90,6 +90,19 @@ public MimeMessage prepareEmailMessage(String recipients, String mapNotification
...
+        }
+       if (sender.isEmpty()) {
```
e-pettersson-ericsson: Use StringUtils.isEmpty to avoid NullPointer in these if statements

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
navya108: yah I have tried putting upon emailSender.prepareEmailMessage but still facing the error.
Actually in EmailSenderTest class its mentioned previously javadocs that 
 /*
     * prepareEmail is not possible to test due too much hidden Spring mumbo jumbo doing things in
     * background.
     */
so I think should remove the testcase.

e-pettersson-ericsson: I took a look at the code again, and it seems very difficult to mock the MimeMessage and MimeMessageHelper... Since the default values doesn't get added until we actually use the prepareEmail method this test is rendered useless. I say remove this unit test and instead keep the cucumber functional test which actually sends an email with the correct subject + sender...

src/main/java/com/ericsson/ei/notifications/EmailSender.java
navya108: okay

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
e-pettersson-ericsson: I agree with you, remove this test case then.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -95,10 +96,10 @@ public MimeMessage prepareEmailMessage(String recipients, String mapNotification
      * at application.proporties are empty.
      * */
     private void setDefaultValuesEmailSenderSubject() {
-       if (subject.isEmpty()) {
+       if (StringUtils.isEmpty(subject) == true) {
             subject = "Email Subscription Notification";
         }
-       if (sender.isEmpty()) {
+       if (StringUtils.isEmpty(sender) == true) {
```
e-pettersson-ericsson: No need for the "== true". The StringUtils.isEmpty(sender) gives a boolean result which can be used as is, in the if statement.

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
```diff
@@ -18,10 +18,12 @@
 import javax.mail.internet.MimeMessage;
 
 import org.junit.Test;
+import static org.junit.Assert.*;
 import org.junit.runner.RunWith;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
+import org.springframework.beans.factory.annotation.Value;
```
e-pettersson-ericsson: It looks like there are unused imports in this file - remove them.

src/main/java/com/ericsson/ei/notifications/EmailSender.java
navya108: Right now its not needed , I just kept it considering the situation (if in case the fields email.sender and email.subject) are not present in application.proporties

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
e-pettersson-ericsson: Codacy Issue found: [Unused import - org.springframework.beans.factory.annotation.Value.](https://app.codacy.com/manual/eiffel-intelligence-maintainers/eiffel-intelligence/pullRequest?prid=4891534)

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
```diff
@@ -18,10 +18,12 @@
 import javax.mail.internet.MimeMessage;
 
 import org.junit.Test;
+import static org.junit.Assert.*;
```
e-pettersson-ericsson: Codacy Issue found: [Avoid unused imports such as 'org.junit.Assert'](https://app.codacy.com/manual/eiffel-intelligence-maintainers/eiffel-intelligence/pullRequest?prid=4891534)

src/test/java/com/ericsson/ei/notifications/EmailSenderTest.java
e-pettersson-ericsson: Codacy Issue found: [Avoid unused imports such as 'org.springframework.beans.factory.annotation.Value'](https://app.codacy.com/manual/eiffel-intelligence-maintainers/eiffel-intelligence/pullRequest?prid=4891534)

src/main/java/com/ericsson/ei/notifications/EmailSender.java
```diff
@@ -48,17 +49,17 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(EmailSender.class);
 
     @Getter
-    @Value("${email.sender}")
+    @Value("${email.sender:noreply@domain.com}")
     private String sender;
 
     @Getter
-    @Value("${email.subject}")
+    @Value("${email.subject:Email Subscription Notification}")
```
tobiasake: Is default subject really needed here, since its defaulted later in the code.?

PR #444, opened 2020-02-26
fix by replacing id with _id

saif-ericsson:
### Applicable Issues
Closes https://github.com/eiffel-community/eiffel-intelligence/issues/441

This is because Eiffel Intelligence doesn't find the resulting aggregation from the database.
The key to search for should be "_id" not "id" - since we can't expect all and every custom rule set to put an "id" in the root of the aggregation.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/src/main/java/com/ericsson/ei/queryservice/ProcessAggregatedObject.java
+++ b/src/main/java/com/ericsson/ei/queryservice/ProcessAggregatedObject.java
@@ -72,7 +72,7 @@ public class ProcessAggregatedObject {
      * @return ArrayList
      */
     public ArrayList<String> getAggregatedObjectByTemplateName(String templateName) {
-        String queryString = "{\"id\": /.*" + templateName + "/}";
+        String queryString = "{\"_id\": /.*" + templateName + "/}";
         MongoQuery query = new MongoStringQuery(queryString);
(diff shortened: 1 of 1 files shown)


Christoffer-Cortes:
A test could be to use a custom rule that doesn't extract any information from the start event so that just the general aggregation information is in the object. Then try to fetch it using the start event id. That should be successful with this fix.

e-pettersson-ericsson:
> SAIF: If there is one, need to be improved

Just to be clear then: have you tested this functionality locally/manually? Since I don't see any tests affected in this PR that is why I'm asking, to make sure this change doesn't break anything unforeseen.

saif-ericsson:
> A test could be to use a custom rule that doesn't extract any information from the start event so that just the general aggregation information is in the object. Then the test aggregation should fetch the object using the start event id. That should be successful with this fix.

saif-ericsson:
> > A test could be to use a custom rule that doesn't extract any information from the start event so that just the general aggregation information is in the object. Then the test aggregation should fetch the object using the start event id. That should be successful with this fix.

I have created a test for for the two untested methods that will cover this comment as well.

Issue #399, opened 2019-11-26
Escape special characters in subscription conditions?


e-pettersson-ericsson:
### Description
Is it possible to escape special characters when writing a condition in a subscription? 
Let's say we have users who want to write a condition where `X == 'something/somethingElse'` ?

Do we need to escape those characters within the JMESPath expression? What characters are allowed? [The documentation](https://github.com/eiffel-community/eiffel-intelligence/blob/master/wiki/subscriptions.md#writing-requirements-and-conditions) needs to be updated with this information. Perhaps the front-end docs also?

### Motivation
Users might want to match objects in the aggregation which contains special characters.
They need answers to: 
* What characters are allowed to use in the JMESPath condition?
* Is it possible to escape special characters?
    * If so, how to do it?

### Benefits
More varied conditions can be written. Clear examples can help users get going faster.

### Possible Drawbacks
None that I can think of.

a-breid-ericsson:
I dont know exactly how special characters works here, but if they break something on the way a possibility is to url encode the string before sending it to back-end and url decode the string before using it. There should be test setup with subscriptions using such conditions.

e-pettersson-ericsson:
An update: I've only tested that it is possible to use / in the JMESPath conditions but I can't really say for other special characters ...

saif-ericsson:
Question: Regarding escaping the special characters, are you talking about the limitation of JEMSPath or limitation of code before calling to JMESPath...?

e-pettersson-ericsson:
I think from a user perspective, the end result is the same. They need to know which (if any) special characters are not allowed to include in their subscription conditions. (If JMESPath expressions have these limitations or if i's somehtign extra we've added on top in Eiffel Intelligence (don't know why we would?) it doesn't really matter. 

I think what I wanted when I wrote this issue was: if some characters are NOT allowed, i.e. because they will be url-encoded or parser etc, this needs to be documented properly. When I wrote the issue I was looking specifically for the / being allowed and had to test it to make sure it was possible to use. It could be that we do not have limitations, i.e. we allow any kind of String in the subcriptions condition, but then we need to clarify this in the documentation: what kind of syntax is expected.

e-pettersson-ericsson:
And if some special characters are not allowed, is it possible to escape this in some JMESPath syntax or is it not possible to use them at all?

saif-ericsson:
[...]

PR #450, opened 2020-04-04
documentation update for subscription

saif-ericsson:
Adding JMESPath link in subscription documentation
#399
### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/wiki/subscriptions.md
+++ b/wiki/subscriptions.md
@@ -44,7 +44,7 @@ This _id serves as an index in Mongo DB to speed up searches in the database.
 The below aggregation is based on the [Artifact rules](../src/main/resources/rules/ArtifactRules-Eiffel-Agen-Version.json).
 
 It should be noted that requirements in subscriptions are written 
-by using [JMESPath](https://jmespath.org/) language.
+by using [JMESPath](https://jmespath.org/) language. And, here is the [JMESPATH specification](https://jmespath.org/specification.html).
 
 **Aggregation from several Eiffel events:**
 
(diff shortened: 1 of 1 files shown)


wiki/subscriptions.md
```diff
@@ -34,7 +34,8 @@ the aggregated object will look like when writing the conditions. The
 requirements inside a subscription should reflect a wanted state of the 
 aggregation, for example: "_Hey Eiffel Intelligence, let me know when my 
 artifact has been published here!_" or "_I want to know when these two products 
-have reached confidence level X_".
+have reached confidence level X_". Requirements in subscriptions are written 
+using [JMESPath](https://jmespath.org/) language.
```
a-breid-ericsson, line 38: I think it would be more clear if "Requirements in subscriptions are written 
using [JMESPath](https://jmespath.org/) language." was it´s own line before the text blob or after the text blob rather then a part of the text blob.

saif-ericsson:
> characters that break the string.

But that is quite standard and need not to be explained. I think then we are like explaining what the strings are and how to write strings.

wiki/subscriptions.md
saif-ericsson: done

m-linner-ericsson:
> 
> 
> > characters that break the string.
> 
> But that is quite standard and need not to be explained. I think then we are like explaining what the strings are and how to write strings.

I just thought it might be good to be explicit and it never hurts to be too clear. However, you are the author so you get to decide :)

PR #468, opened 2020-06-12
Old bindingkey is removed if binding key is changed in configuration

zrddnvt:
### Description of the Change
If binding key is changed in configuration, the old bindingkey is removed from mb queue.

### Benefits
Reduction in conflicts between the binding keys.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: Navyatha Reddi navyatha.reddi@tcs.com, Durga vasaadi durga.vasaadi@tcs.com, Surya Kumari Kolli

### DIFF
--- a/src/main/java/com/ericsson/ei/handlers/RmqHandler.java
+++ b/src/main/java/com/ericsson/ei/handlers/RmqHandler.java
@@ -130,7 +130,7 @@ public class RmqHandler {
     @JsonIgnore
     private AmqpAdmin amqpAdmin;
 
-    @Getter
+    @Setter
     @JsonIgnore
     private RabbitManagementTemplate rabbitManagementTemplate;
 
@@ -164,15 +164,11 @@ public class RmqHandler {
 
     @Bean
     public RabbitManagementTemplate deleteBindingKeys() {
-    	Integer rabbitPort=port;
-    	String rabbitHttp = "https://";
-    	if(port==5672 && host.equals("localhost")){
+    	int rabbitPort=port;
+    	if(port==5672){
     		rabbitPort=15672;
-    		rabbitHttp="http://";
     	}
-        System.out.println("http="+rabbitHttp);
-        System.out.println("rabbitPort="+rabbitPort);
-        rabbitManagementTemplate = new RabbitManagementTemplate(rabbitHttp+host+":"+rabbitPort+"/api/",user,password);
+        rabbitManagementTemplate = new RabbitManagementTemplate("http://"+host+":"+rabbitPort+"/api/",user,password);
         amqpAdmin = new RabbitAdmin(connectionFactory());
         rabbitManagementTemplate.getBindings().stream().filter(
         binding -> binding.getDestination().equals(externalQueue().getName()) && binding.isDestinationQueue())
@@ -180,7 +176,8 @@ public class RmqHandler {
             final String[] bingingKeysArray = splitBindingKeys(bindingKeys);
             for (final String bindingKey : bingingKeysArray) {
 	            if(binding.getRoutingKey().equals(bindingKey)){
-	               System.out.println("Not Deleting the existing binding key : "+bindingKey);
+	               System.out.println("............."+bindingKey);
+	               LOGGER.debug("Not Deleting the existing binding key : "+bindingKey);
 	               break;
 	             }
 	             else{
--- a/src/main/java/com/ericsson/ei/handlers/RmqHandler.java
+++ b/src/main/java/com/ericsson/ei/handlers/RmqHandler.java
@@ -248,7 +248,7 @@ public class RmqHandler {
     // comparing with the stored key and adding the new binding key into the mongoDB.
        for(final Binding bindingKey:AMQPBindingObjectList){
             if(existingBindingsData.contains(bindingKey.getRoutingKey())){
-                LOGGER.info("Binding already present");
+                LOGGER.info("Binding already present in mongoDB");
             }else{
                     BasicDBObject document = new BasicDBObject();
                     document.put("destination",bindingKey.getDestination());
--- a/src/main/java/com/ericsson/ei/handlers/RmqHandler.java
+++ b/src/main/java/com/ericsson/ei/handlers/RmqHandler.java
@@ -22,12 +22,15 @@ import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.amqp.core.AcknowledgeMode;
+import org.springframework.amqp.core.AmqpAdmin;
 import org.springframework.amqp.core.Binding;
 import org.springframework.amqp.core.BindingBuilder;
 import org.springframework.amqp.core.Queue;
 import org.springframework.amqp.core.TopicExchange;
 import org.springframework.amqp.rabbit.connection.CachingConnectionFactory;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
+import org.springframework.amqp.rabbit.core.RabbitAdmin;
+import org.springframework.amqp.rabbit.core.RabbitManagementTemplate;
 import org.springframework.amqp.rabbit.core.RabbitTemplate;
 import org.springframework.amqp.rabbit.core.RabbitTemplate.ConfirmCallback;
 import org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer;
@@ -122,6 +125,14 @@ public class RmqHandler {
     @JsonIgnore
     private SimpleMessageListenerContainer container;
 
+    @Getter
+    @JsonIgnore
+    private AmqpAdmin amqpAdmin;
+
+    @Getter
+    @JsonIgnore
+    private RabbitManagementTemplate rabbitManagementTemplate;
+
     @Bean
     public ConnectionFactory connectionFactory() {
         cachingConnectionFactory = new CachingConnectionFactory(host, port);
@@ -172,12 +183,24 @@ public class RmqHandler {
 
     @Bean
     public List<Binding> bindings() {
+        deleteBindingKeys();
         final String[] bingingKeysArray = splitBindingKeys(bindingKeys);
         final List<Binding> bindingList = new ArrayList<>();
         for (final String bindingKey : bingingKeysArray) {
             bindingList.add(BindingBuilder.bind(externalQueue()).to(exchange()).with(bindingKey));
         }
         return bindingList;
+    }
+
+    public void deleteBindingKeys() {
+        rabbitManagementTemplate = new RabbitManagementTemplate();
+        amqpAdmin = new RabbitAdmin(connectionFactory());
+        rabbitManagementTemplate.getBindings().stream().filter(
+                binding -> binding.getDestination().equals(externalQueue().getName()) && binding.isDestinationQueue())
+                .forEach(binding -> {
+                    LOGGER.debug("Deleting " + binding);
+                    amqpAdmin.removeBinding(binding);
+                });
     }
 
     @Bean
--- a/src/main/java/com/ericsson/ei/handlers/RmqHandler.java
+++ b/src/main/java/com/ericsson/ei/handlers/RmqHandler.java
@@ -18,12 +18,13 @@ package com.ericsson.ei.handlers;
 
 import java.util.ArrayList;
 import java.util.List;
-
+import org.bson.Document;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.amqp.core.AcknowledgeMode;
 import org.springframework.amqp.core.AmqpAdmin;
 import org.springframework.amqp.core.Binding;
+import org.springframework.amqp.core.Binding.DestinationType;
 import org.springframework.amqp.core.BindingBuilder;
 import org.springframework.amqp.core.Queue;
 import org.springframework.amqp.core.TopicExchange;
@@ -43,6 +44,10 @@ import org.springframework.stereotype.Component;
 
 import com.ericsson.ei.listener.EIMessageListenerAdapter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.mongodb.BasicDBObject;
+import com.mongodb.MongoClient;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoDatabase;
 
 import lombok.Getter;
 import lombok.Setter;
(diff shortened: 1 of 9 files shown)


Christoffer-Cortes:
There is an error with that new bean:

Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'rmqHandler': Unsatisfied dependency expressed through field 'rabbitManagementTemplate'; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'deleteBindingKeys' defined in class path resource [com/ericsson/ei/handlers/RmqHandler.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.amqp.rabbit.core.RabbitManagementTemplate]: Circular reference involving containing bean 'rmqHandler' - consider declaring the factory method as static for independence from its containing instance. Factory method 'deleteBindingKeys' threw exception; nested exception is org.springframework.web.client.ResourceAccessException: I/O error on GET request for "http://localhost:15672/api/bindings/": Connect to localhost:15672 [localhost/127.0.0.1] failed: Connection refused (Connection refused); nested exception is org.apache.http.conn.HttpHostConnectException: Connect to localhost:15672 [localhost/127.0.0.1] failed: Connection refused (Connection refused)

src/main/java/com/ericsson/ei/handlers/RmqHandler.java
```diff
@@ -150,6 +162,26 @@ public ConnectionFactory connectionFactory() {
         return cachingConnectionFactory;
     }
 
+    @Bean
+    public RabbitManagementTemplate deleteBindingKeys() {
+    	Integer rabbitPort=port;
```
Christoffer-Cortes, line 167: Why not use primitive type int?

src/main/java/com/ericsson/ei/handlers/RmqHandler.java
```diff
@@ -150,6 +162,26 @@ public ConnectionFactory connectionFactory() {
...
+    	String rabbitHttp = "https://";
+    	if(port==5672 && host.equals("localhost")){
```
Christoffer-Cortes, line 169: What is the logic behind this code block? The API is exposed through the management page usually located at 15672 but if the port is anything other than 5672 we don't attempt to redirect to the website API? Won't this fail unless the user specifially configures the API url?

zrddnvt: Hi,Actually for removing old binding keys we have gone through some websites on google, and found out the RabbitManagementTemplate to remove the old binding keys and used for implementation. So as part of it we are facing issues with RabbitManagementTemplate in test class. So we tried to solve it with this kind of approach, just for testing purpose.

src/main/java/com/ericsson/ei/handlers/RmqHandler.java
```diff
@@ -150,6 +162,33 @@ public ConnectionFactory connectionFactory() {
         return cachingConnectionFactory;
     }
 
+    @Bean
+    public RabbitManagementTemplate deleteBindingKeys() {
+    	int rabbitPort=port;
+    	if(port==5672){
+    		rabbitPort=15672;
+    	}
+        rabbitManagementTemplate = new RabbitManagementTemplate("http://"+host+":"+rabbitPort+"/api/",user,password);
```
tobiasake, line 171: What happens if un-secure http is disabled and only https is available in target environment? 
Then this implementation will not work, I guess.

raja-maragani, line 171: RabbitManagementTemplate will use the http calls to connect with RabbitMQ and RabbitTemplate uses the amqp. both implementations are different. so better to change the new implementation to amqp based.

zrddnvt: As per raja's comment, we have investigated for the other way of getting the binding keys form the RabbitMq,but rather that RabbitManagmentTemplate/Client we couldn't find any way to fetch the list of bindings using current EI implementation (Spring AMQP). Also digged into the code level referring the RabbitTemplate, channels and connectionFactory but there is no such functionality to get the bindingKeys.

So other way of approach for this implementation is using files, i.e to store the binding keys into the file/List and check with the existing bindingkeys, add or update the file with the required binding keys on EI start. The file gets updated when the binding configuration is changed.With this approach the manually created bindings cannot be retrieved which in turn cannot be deleted from the code.
Please suggest if this can be implemented in any other way.

tobiasake: How is other Eiffel component handling restart of Eiffel components with new bindingKey ?

tobiasake: Can't EI create a new queue with the new BindingKey name as suffix to the queue name and let the administrator choose to do with old queue and old event data, if it should be kept or removed.
EI starts to consume messages from new queue with new binding key(s) when started with new binding keys.
By doing that we don't wipe any persistent data or what it can cause.

tobiasake: Or if the administrator want the old queue and old bindingkeys to be removed automatically, set durable queue flag to False and old queue+bindingkeys will be removed automatically at shutdown and when EI starts a new queue will be created with new bindingkey(s), if I understand it correctly.

zrddnvt: > Or if the administrator want the old queue and old bindingkeys to be removed automatically, set durable queue flag to False and old queue+bindingkeys will be removed automatically at shutdown and when EI starts a new queue will be created with new bindingkey(s), if I understand it correctly.

yah, but apart from setting durable queue flag, the actually issue is still staying i.e old binding keys which are not in use are staying back in mb queue. So is there any other way programmatically to remove old binding keys in mb queue.

tobiasake: Have you looked into Queue TTL:
https://www.rabbitmq.com/ttl.html#queue-ttl

So queue is removed by RabbitMq itself after expire time.
This involves adding a new rabbitmq property to application.properties.

zrddnvt: yah,but accordingly as far as I know, with the same queue there will be different binding keys, so  binding keys which are not in use must be removed, so remaining binding keys which are in use with that queue will stay. If TTL is used entire queue will be in effect irrespective of binding keys. so we are trying to get the binding keys from mb queue. Correct me if I am wrong.

tobiasake: I think the only solution is to bind and unbind the queue when EI starts and shutsdowns.
Since EI is not aware of previous binding keys used in other/previous running EI instances.
Documentation
https://pika.readthedocs.io/en/latest/modules/channel.html?highlight=bind#pika.channel.Channel.queue_unbind

If I understand what you trying to achieve.

I thinks its not safe/secure to let next EI instance to figure out what previously binding key was used and remove it.

Other option is what I commented about in previous comment, create a new queue with the new binding key and let rabbitmq remove the old queue+bindingkey with Queue TTL flag.

raja-maragani: > How is other Eiffel component handling restart of Eiffel components with new bindingKey ?

This binding change will happen from some of the Eiffel components, but rebind is not implemented in Eiffel components.

zrddnvt: Had a discussion regarding the implementation way and the proposed solution suggested in the meeting is to use mongodb collection to store the binding keys and use the collection for removing the binding keys and updating the new binding keys.
Will start the implementation using mongo db with collection name "BindingKeys".

e-pettersson-ericsson: If we see this behavior on latest version of Eiffel Intelligence (and I think it is the same) we will need to transfer this fix also to the master branch, which will be included in an upcoming 4.x release (when the API versioning has been added)

tobiasake: Isn't "binding-keys" better collection name,, then its easier to administrate MongoDb if using mongo client command and typing the collection name in mongo shell.
Its easier to type only lower letters in the shell.

tobiasake: This collection need to be unique also for each EI instance if same mongodb is used, so collection name need to be configurable from application.config.
So we get unique collection name per instance, if same MongoDb is used.
Same MongoDb is used by tests and also in Easy2Use, maybe in other setups as well.

src/main/java/com/ericsson/ei/handlers/RmqHandler.java
```diff
@@ -108,9 +123,17 @@
     @Setter
     @Value("${rabbitmq.consumerName}")
     private String consumerName;
+    
+    @Getter
+    @Value("${spring.data.mongodb.database}")
+    private String bindingKeysDataBaseName;
```
tobiasake, line 129: I guess its same database for whole EI instance, I think its better to name variable just "dataBaseName", so we avoid confusing it with different databse.
I guess we want to achieve one database per EI instance, if several EI instances using same MongoDb instance.
So we have following structure:
EIInstanceDb1:
   - SubscriptionCollection
   - AggregatedObjectCollection
   - MatchedSubscriptionCollection
   - BindingKeyCollection
   - and so on, migh have missed a collection that EI uses

EIInstanceDb2:
   - SubscriptionCollection
   - AggregatedObjectCollection
   - MatchedSubscriptionCollection
   - BindingKeyCollection
   - and so on

EIInstanceDb3:
   - SubscriptionCollection
   - AggregatedObjectCollection
   - MatchedSubscriptionCollection
   - BindingKeyCollection
   - and so on

I guess the BindingKey Collection name can be hardcoded in a Constant library, so we have the name string defined in one place. I guess we have this for other collection names already.

src/main/java/com/ericsson/ei/handlers/RmqHandler.java
```diff
@@ -171,16 +198,59 @@ Binding binding() {
     }
 
     @Bean
-    public List<Binding> bindings() {
+    public List<Binding> bindings(){
         final String[] bingingKeysArray = splitBindingKeys(bindingKeys);
```
tobiasake, line 202: "bingingKeysArray" -> should be named : "bindingKeysArray"

src/main/java/com/ericsson/ei/handlers/RmqHandler.java
```diff
@@ -171,16 +198,59 @@ Binding binding() {
...
         final List<Binding> bindingList = new ArrayList<>();
         for (final String bindingKey : bingingKeysArray) {
             bindingList.add(BindingBuilder.bind(externalQueue()).to(exchange()).with(bindingKey));
         }
+        deleteBindings(bingingKeysArray,bindingList);
         return bindingList;
     }
 
-    @Bean
+    private void deleteBindings(String[] bingingKeysArray, List<Binding> bindingList){
```
tobiasake, line 211: Can you put better variable names on input parameter "String[] bingingKeysArray, List<Binding> bindingList" , so its easier to understand which bindings is new bindings and what difference between those.
Or write Java doc.
Example of names , e.g:
"newBindingKeysArray"
"AMQPBindingObjectList"

Or something similar.

src/main/java/com/ericsson/ei/handlers/RmqHandler.java
```diff
@@ -171,16 +198,59 @@ Binding binding() {
...
+                        mongoDBHandler.dropDocument(bindingKeysDataBaseName, bindingKeyCollectionName, condition);
+                    }
+                    else{
+                        // storing the existing key into an array.
+                        existingBindingsData.add(mongoDbBindings);
+                    }
+            }
+        }
+        // comparing with the stored key and adding the new binding key into the mongoDB.
+        for(final Binding bindingKey:bindingList){
```
tobiasake, line 238: This part of the code that store newBindingKeys into MongoDb, could be put to an own private function where the function name says what it do:

private void storeNewBindingKeys(ArrayList<> AMQPBindingObjectList)

or something.

zrddnvt:
> And tests is missing.

Tried implementing the test scenario for the binding keys stored in mongodb. Please help me out in any changes or any test cases to be added.

tobiasake:
> > And tests is missing.
> 
> Tried implementing the test scenario for the binding keys stored in mongodb. Please help me out in any changes or any test cases to be added.
I see  that you are using some RabbitAdmin library,, I guess that is a RabbitMq API specific library, does that work with normal AMQP standards and Qpid which is the embedded AMQP server that is used for functional tests.
If RabbitAdmin only works with RabbitMq and not AMQP in general, then I guess this is only possible to test as an integrationTests. I think we have the integrationTests in the EI-Frontend, which tests both backend and frontend .. ?
Someone can correct me if I am wrong!
From that test framework we use docker-compose and RabbitMq AMQP server.
In that test framework  we could add integration test which tests start EI-backend with one bindingkey and then restart EI-Backend with new bindingkey and check RabbitMq queue if bindinkey has changed and also check that MongoDB bindingKey collection has removed the old bindingKey and newBindingkey has been added.
What is your ideas about testing this? @e-pettersson-ericsson  @Christoffer-Cortes

Christoffer-Cortes:
> > > And tests is missing.
> > 
> > 
> > Tried implementing the test scenario for the binding keys stored in mongodb. Please help me out in any changes or any test cases to be added.
> > I see  that you are using some RabbitAdmin library,, I guess that is a RabbitMq API specific library, does that work with normal AMQP standards and Qpid which is the embedded AMQP server that is used for functional tests.
> > If RabbitAdmin only works with RabbitMq and not AMQP in general, then I guess this is only possible to test as an integrationTests. I think we have the integrationTests in the EI-Frontend, which tests both backend and frontend .. ?
> > Someone can correct me if I am wrong!
> > From that test framework we use docker-compose and RabbitMq AMQP server.
> > In that test framework  we could add integration test which tests start EI-backend with one bindingkey and then restart EI-Backend with new bindingkey and check RabbitMq queue if bindinkey has changed and also check that MongoDB bindingKey collection has removed the old bindingKey and newBindingkey has been added.
> > What is your ideas about testing this? @e-pettersson-ericsson  @Christoffer-Cortes

Christoffer-Cortes:
Sorry about the closing. =P Github reply/comment can be a bit confusing at times.
Anyway, either an integration test with docker components in the frontend if there is a way to programmatically restart a docker component... or a functionaltest and do like in scaling and failover test and start/close spring application from the code.

SantoshNC68:
Merging the PR.

PR #474, opened 2020-09-03
EIB: Jenkins job not getting triggered after version uplift to 2.222.3

Surya-Kolli:
### Description of the Change
The jenkins jobs were not getting triggered as part of jenkins version uplift(2.222.3), which involves CSRF protection using jenkins crumbs.
The rest url post requests were getting failed because of invalid crumb request, so changed the url parsing and added the context path(/jenkins).

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: K.Suryakumari suryakumari.kolli@ericsson.com

### DIFF
--- a/src/main/java/com/ericsson/ei/notifications/JenkinsCrumb.java
+++ b/src/main/java/com/ericsson/ei/notifications/JenkinsCrumb.java
@@ -95,7 +95,9 @@ public class JenkinsCrumb {
      */
     private URL buildJenkinsCrumbUrl(String url) throws MalformedURLException {
         String baseUrl = urlParser.extractBaseUrl(url);
-        URL jenkinsCrumbUrl = new URL(baseUrl + JENKINS_CRUMB_ENDPOINT);
+        String contextPath = urlParser.extractContextPath(url).split("/")[1];
+        String jenkinsBaseUrl = String.format("%s/%s", baseUrl, contextPath);
+        URL jenkinsCrumbUrl = new URL(jenkinsBaseUrl + JENKINS_CRUMB_ENDPOINT);
         return jenkinsCrumbUrl;
     }
 }
--- a/src/main/java/com/ericsson/ei/notifications/JenkinsCrumb.java
+++ b/src/main/java/com/ericsson/ei/notifications/JenkinsCrumb.java
@@ -96,8 +96,13 @@ public class JenkinsCrumb {
     private URL buildJenkinsCrumbUrl(String url) throws MalformedURLException {
         String baseUrl = urlParser.extractBaseUrl(url);
         String contextPath = urlParser.extractContextPath(url).split("/")[1];
-        String jenkinsBaseUrl = String.format("%s/%s", baseUrl, contextPath);
-        URL jenkinsCrumbUrl = new URL(jenkinsBaseUrl + JENKINS_CRUMB_ENDPOINT);
+        URL jenkinsCrumbUrl;
+        if (contextPath.equals("job")) {
+            jenkinsCrumbUrl = new URL(baseUrl + JENKINS_CRUMB_ENDPOINT);
+        } else {
+            String jenkinsBaseUrl = String.format("%s/%s", baseUrl, contextPath);
+            jenkinsCrumbUrl = new URL(jenkinsBaseUrl + JENKINS_CRUMB_ENDPOINT);
+        }
         return jenkinsCrumbUrl;
     }
 }
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.github.ericsson</groupId>
     <artifactId>eiffel-intelligence</artifactId>
-    <version>2.2.3</version>
+    <version>2.2.4</version>
     <packaging>war</packaging>
 
     <parent>
--- a/src/main/java/com/ericsson/ei/notifications/UrlParser.java
+++ b/src/main/java/com/ericsson/ei/notifications/UrlParser.java
@@ -104,7 +104,7 @@ public class UrlParser {
      * @throws MalformedURLException
      * @throws URISyntaxException
      */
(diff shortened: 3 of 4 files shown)


src/main/java/com/ericsson/ei/notifications/JenkinsCrumb.java
```diff
@@ -95,7 +95,9 @@ public JsonNode fetchJenkinsCrumb(String encoding, String url)
      */
     private URL buildJenkinsCrumbUrl(String url) throws MalformedURLException {
         String baseUrl = urlParser.extractBaseUrl(url);
-        URL jenkinsCrumbUrl = new URL(baseUrl + JENKINS_CRUMB_ENDPOINT);
+        String contextPath = urlParser.extractContextPath(url).split("/")[1];
+        String jenkinsBaseUrl = String.format("%s/%s", baseUrl, contextPath);
```
tobiasake, line 99: Is this valid for valid for all Jenkins versions? Does this work with previous Jenkins versions as well?
What has changed between previous Jenkins version bad new Jenkins version for handling Crumb Auth?

tobiasake: I checked the subscription,, I guess url is
http://<JenkinsHost:port>/job/<JobName>/job/<branch>/build
Context-path refers in current code base rest api entrypoint path to trigger the jenkins build.
So code need to rewritten so it might be clear.
And I guess you want handle the Tomcat Context-path that Tomcat adds in the middle.
That conext-path is more the real context path.
So code need to re-written a bit to in a smart way figure out the real context-path that is added by Tomcat in url address and in smart way extract where does the Jenkins url/entrypoint path starts.

Is this the case you try to solve?
http://<JenkinsHost:port>/<web server application context path>/job/<JobName>/job/<branch>/build

src/main/java/com/ericsson/ei/notifications/JenkinsCrumb.java
```diff
@@ -95,7 +95,9 @@ public JsonNode fetchJenkinsCrumb(String encoding, String url)
      */
     private URL buildJenkinsCrumbUrl(String url) throws MalformedURLException {
         String baseUrl = urlParser.extractBaseUrl(url);
-        URL jenkinsCrumbUrl = new URL(baseUrl + JENKINS_CRUMB_ENDPOINT);
+        String contextPath = urlParser.extractContextPath(url).split("/")[1];
```
tobiasake, line 98: This will break functionality when its deployed with no web-server context-paths, then you will get first part of Jenkins entrypoint paths as context-path, which we do in several deployments, e.g. in Docker and Kubernetes and Integration-Tests. I think also in frontend Integrations-tests.
So you need to make a smarter solution that support with and without web-server context-paths.

tobiasake (changes requested):

src/main/java/com/ericsson/ei/notifications/JenkinsCrumb.java
a-breid-ericsson: I agree with Tobias here, this is a fix for an Ericsson specific problem that will break the functionality for everyone else.

tobiasake: One suggestion to solve this is to introduce context-path tagging in the Jenkins url.
If we have this full url Jinkens jobb:
http://some_host:12345/jenkins-context-path/job/hello-word/build

In subscription Meta field user need to replace the slahses with some other character/tag, for marking that this is context-path, example if we can use '|' as tag/context-path marker:
http://some_host:12345|jenkins-context-path/job/hello-word/build

In EI code we can then check if url has "|" character , if it has that character in the first part of url after port-number, then its a context-path. Based on the we can devide it up to url variables in java code, and replace '|' with a '/' :
String jenkinsBaseUrl="http://some_host:12345/jenkins-context-path"
String jenkinsJobBuildUrlPath=/job/hello-word/build

Now you can make a CSRF Crumb authentication with help of "jenkinsBaseUrl" variable.
To trigger the jenkins job after CSRF Crumb authentication, then contactenate these to variables to full url and make the HTTP request for triggering the job.


I think this change is backward compatible and it will work with and without context-path.

Documentation in Frontend and in the Web-UI need to be made, so its clear to user how it works with context-paths Jenkins url address.

Surya-Kolli: Yes this approach sounds good, and we also assumed the same like but checking with the keyword /job after base url depending on this we can differentiate between with contextPath and without contextPath assuming the contextPath lies in between the baseUrl and /job.
But here we have few doubts like whether the context path can be "/jenk/ins" can it have / also , if this is the case we need to handle such.
So currently we are checking jenkins rules and url standards and we will try once in local the suggestion proposed.

Surya-Kolli: @Tobias can we proceed with with the above solution assuming the jenkins url will be:
http://some_host:12345/jenkins-context-path/job/hello-word/build

considering the contextpath lies between base url and /job

tobiasake: Ok, I guess its an OK solution. I am no super expert on which Jenkins url entrypoints paths.
As long test goes successful, then its a OK solution.
I guess we have some Subscription Jenkins Triggering tests in System tests in EI Frontend git repository as well, those tests need to be executed as well, you can do that manually on your local computer or something.

Surya-Kolli: Ok Tobias , will test the chages

tobiasake (approved):
Looks good.

tobiasake:
We need this change on master branch as well, can someone do the same update on master branch as well?

a-breid-ericsson (approved):

SantoshNC68 (approved):

PR #508, opened 2021-09-16
Fixing missing events issues

vrpanchal:
### Applicable Issues
Fixing missing events issues when multiple events were getting lost due to lock on mongoDB.

### Description of the Change
Fixing missing events issues when multiple events were getting lost due to lock on mongoDB.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: Vishalkumar Panchal vr.panchal@tcs.com


src/main/java/com/ericsson/ei/controller/RuleTestControllerImpl.java
```diff
@@ -85,8 +86,13 @@
             final HttpServletRequest httpRequest) {
         if (testEnabled) {
             try {
-                String aggregatedObject = ruleTestService.prepareAggregatedObject(
-                        new JSONArray(body.getListRulesJson()), new JSONArray(body.getListEventsJson()));
+                String aggregatedObject = StringUtil.EMPTY_STRING;
+                try {
```
SantoshNC68, line 90: why this was written inside another try?

src/main/java/com/ericsson/ei/controller/RuleTestControllerImpl.java
```diff
@@ -85,8 +86,13 @@
...
+                        aggregatedObject = ruleTestService.prepareAggregatedObject(
+                                new JSONArray(body.getListRulesJson()), new JSONArray(body.getListEventsJson()));
+                } catch (Exception e) {
+                    e.printStackTrace();
```
SantoshNC68, line 94: donot print stacktrace, use loggers to log them

src/main/java/com/ericsson/ei/controller/RuleTestControllerImpl.java
vrpanchal: It is require, as in sub try/catch we have to catch Exception and we are not returning anything. If we use one try/catch then we need to add finally block to return ResponseEntity, and Finally block is considered to be cleanup block, return is not generally expected in it. Hence, nested try/catch is require here.

src/main/java/com/ericsson/ei/controller/RuleTestControllerImpl.java
vrpanchal: Corrected.

